PR #103, opened 2026-02-11
feat: Text Field Component

SamFanvue:
## Summary

completes ENG-7425
## Changes

- Add TextField component with tests and stories

## Screenshots / Storybook

image

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -47,7 +47,7 @@ export const TextField = React.forwardRef<HTMLInputElement, TextFieldProps>(
 
     return (
       <div className={cn("flex flex-col", fullWidth ? "w-full" : "", className)}>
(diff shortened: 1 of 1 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,125 @@
...
+  ) => {
+    const generatedId = React.useId();
+    const inputId = id || generatedId;
+    const containerId = `${inputId}-container`;
+    const helperTextId = `${inputId}-helper`;
+    const [isFocused, setIsFocused] = React.useState(false);
+
+    return (
+      <div className={cn("flex flex-col", fullWidth ? "w-full" : "", className)}>
+        {label && (
```
simonstaton: For instances where we don't have a label we should set aria-label for accessibility/screen-readers

PR #115, opened 2026-02-12
fix(a11y,seo): resolve Lighthouse accessibility and SEO audit failures

simonstaton:
- Add aria-label to all IconButton usages in showcase (button-name)
- Add dev warning to IconButton when no accessible name is provided
- Replace opacity-50 with text-body-200 on icon labels for sufficient
  color contrast (4.5:1 ratio)
- Add TabsContent to single-tab showcases to fix invalid aria-controls
  references (aria-valid-attr-value)
- Add aria-label to Slider missing an accessible name
  (aria-input-field-name)
- Increase Pagination dot touch targets from 6-8px to 24px minimum
  (target-size)
- Wrap showcase content in <main> element (landmark-one-main)
- Add meta description to index.html (meta-description)
- The Checkbox component rendered without a `label`, `aria-label`, or
`aria-labelledby` prop produces a button with no accessible name. Add a
development-only console.warn matching the pattern used by TextField and
IconButton, and add aria-label to the bare Checkbox in the showcase.

### DIFF
--- a/.github/workflows/showcase-preview.yml
+++ b/.github/workflows/showcase-preview.yml
@@ -1,0 +1,79 @@
+name: Showcase Preview
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, closed]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: showcase-preview-${{ github.event.number }}
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    name: Deploy Preview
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -116,7 +116,7 @@ function SliderShowcase() {
       />
       <Slider defaultValue={[20, 80]} label="Price Range" minLabel="$0" maxLabel="$1000" />
```
simonstaton: These changes in App.tsx are pretty pointless, just so I can get the report passing when testing things

PR #147, opened 2026-02-17
feat: search input component. Includes debounce and minChar features

SamFanvue:
## Summary

closes ENG-7429

## Changes

- search input component
- debounce and minChar features


SamFanvue:
there are no focus rings on the inputs according to figma

PR #152, opened 2026-02-17
fix: Fix clear icon bugs

SamFanvue:
## Changes

- Fix clear button not appearing on text area
- Replace button with IconButton
- Hide webkit clear input button


simonstaton (commented):
A few other nits inline around `useCallback` and `textareaStyle`, plus a question on the `onChange`-not-called-on-clear contract change.

PR #160, opened 2026-02-19
fix: override browser autofill background on inputs

simonstaton:
Screenshot 2026-02-18 at 20 57 27

## Summary
- Adds `:-webkit-autofill` styles to `theme.css` that prevent the browser from painting a white/blue background on autofilled inputs
- Uses `transition: background-color 9999s` to delay the forced background indefinitely and `-webkit-text-fill-color` to preserve the theme's text color
- Applies to both `input` and `textarea` elements

## Test plan
- [ ] Open a form with a password or email field in dark mode
- [ ] Let the browser autofill the field
- [ ] Verify the input background stays transparent (no white/blue flash)
- [ ] Verify the autofilled text color matches the theme

### DIFF
--- a/src/styles/theme.css
+++ b/src/styles/theme.css
@@ -19,11 +19,13 @@
 }
 
 @theme {
-  --breakpoint-sm: 850px;
-  --breakpoint-md: 1280px;
+  --shadow-box-shadow-med: 4px #0000000d 0px 2px 0px;
+  --shadow-box-shadow-small: 2px #0000000d 0px 2px 0px;
+  --shadow-nav-box-shadow: 12px #0000000d 0px -2px 0px;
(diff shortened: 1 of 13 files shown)


src/styles/theme.css
```diff
@@ -6,6 +6,16 @@
...
   }
+
+  input:-webkit-autofill,
+  input:-webkit-autofill:hover,
+  input:-webkit-autofill:focus,
+  textarea:-webkit-autofill,
+  textarea:-webkit-autofill:hover,
+  textarea:-webkit-autofill:focus {
+    -webkit-text-fill-color: var(--color-body-100);
+    transition: background-color 9999s ease-in-out 0s;
```
simonstaton: Was skeptical about this line but its a legit fix

PR #423, opened 2026-04-10
feat(item-card): adds new item card component

PanchoBubble:
## Summary
Adds new card props to modify layout, make it a clickable

## Changes
```
  /** Layout direction of the card content. @default "vertical" */
  direction?: CardDirection;
  /** HTML element to render. Use "button" for clickable cards. @default "div" */
  as?: CardAs;
  ```       

## Screenshots / Storybook

Screenshot 2026-04-13 at 09 21 10
Screenshot 2026-04-13 at 09 21 15
Screenshot 2026-04-13 at 09 20 56
Screenshot 2026-04-13 at 09 21 03

### CODE REVIEW

**Comment by SamFanvue** [comment]
@PanchoBubble this already exists in the internal library

**Comment by PanchoBubble** [comment]
> @PanchoBubble this already exists in the internal library

@SamFanvue 
can't really use the current component for it, I'll adjust existing card so it can have horizontal layout, as prop, and have some subcomponents and stuff so it can be composed on the pandora side then if you are okay with that

**Comment by SamFanvue** [comment]
Yep that's fine. The component in the internal library can be adjusted however we need.

File: src/App.tsx [hunk 1]
```diff
@@ -4166,6 +4166,71 @@ function CardDemo() {
...
+            </div>
+            <Badge
+              variant="default"
+              leftDot={false}
+              className="typography-semibold-badge h-5 text-content-primary"
+            >
+              MESSAGING
+            </Badge>
+          </div>
+        </div>
```
PanchoBubble, line 4232: should I create more components for all these? @SamFanvue

PR #440, opened 2026-04-16
feat(table): adds settigns variant

PanchoBubble:
## Summary

Screenshot 2026-04-16 at 17 29 30


SamFanvue:
This definitely doesn't need to be a variant. Can and should be built in pandora.

PR #454, opened 2026-04-29
feat(CyclingText): add component for cycling strings with slide animation

simon-fanvue:
## Summary

A rotating text animation. Sits inside divs, spans, buttons, or fake input placeholders. Takes an array of strings and a cycle delay; on tick, the visible item slides out while the next slides in.

- prefers-reduced-motion hard-swap fallback
- pause when document is hidden, so the cycle doesn't catch up after a tab switch
- ResizeObserver-driven width measurement so font-load and parent resize stay in sync
- aria-live="polite" on the visible label, aria-hidden on the sizing and incoming layers
- paused prop for external control (focus, off-screen, etc.)

Sizing has two modes: "longest" (default) reserves space for the longest item to avoid width jitter; "current" animates the wrapper width between items.

## Changes

-

## Checklist

- [ ] TypeScript compiles (`pnpm typecheck`)
- [ ] Linter passes (`pnpm lint`)
- [ ] Tests pass (`pnpm test`)
- [ ] New/updated components have stories
- [ ] New/updated components have accessibility tests
- [ ] New/updated components have forwardRef unit tests
- [ ] New/updated components have className chaining unit tests
- [ ] Exported in `src/index.ts` (if consumable by vendors)
- [ ] Figma link added to story `design` parameter (if applicable)
- [ ] Does not have barrel file `src/YourComponent/index.ts`

## Screenshots / Storybook

image

https://github.com/user-attachments/assets/eb90d5a4-924e-41b3-8796-0be51d4e5cf0

### DIFF
--- a/src/components/CyclingText/CyclingText.stories.tsx
+++ b/src/components/CyclingText/CyclingText.stories.tsx
@@ -55,7 +55,7 @@ export const InsideAButton: Story = {
 
 export const InsideAStatusRow: Story = {
   render: (args) => (
-    <div className="text-content-tertiary inline-flex items-center gap-2">
+    <div className="inline-flex items-center gap-2 text-content-tertiary">
       <SpinnerIcon className="size-4 animate-spin" />
       <CyclingText {...args} />
     </div>
@@ -99,23 +99,18 @@ export const FakePlaceholder: Story = {
           onChange={(e) => setValue(e.target.value)}
           onFocus={() => setFocused(true)}
           onBlur={() => setFocused(false)}
-          className="border-border-default bg-bg-primary text-content-primary w-full rounded-md border px-3 py-2 outline-none"
+          className="w-full rounded-md border border-border-default bg-bg-primary px-3 py-2 text-content-primary outline-none"
         />
         {showPlaceholder && (
           <CyclingText
             {...args}
-            className="text-content-tertiary pointer-events-none absolute inset-y-0 left-3 flex items-center"
+            className="pointer-events-none absolute inset-y-0 left-3 flex items-center text-content-tertiary"
           />
         )}
       </div>
     );
   },
   args: {
-    items: [
-      "Search creators…",
-      "Find a fan…",
-      "Look up a transaction…",
-      "Browse posts…",
-    ],
+    items: ["Search creators…", "Find a fan…", "Look up a transaction…", "Browse posts…"],
   },
 };
--- a/src/components/CyclingText/CyclingText.stories.tsx
+++ b/src/components/CyclingText/CyclingText.stories.tsx
@@ -1,0 +1,121 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import { useState } from "react";
+import { Button } from "../Button/Button";
+import { SpinnerIcon } from "../Icons/SpinnerIcon";
+import { CyclingText } from "./CyclingText";
+
+const meta = {
+  title: "Components/CyclingText",
+  component: CyclingText,
+  parameters: {
+    layout: "centered",
+  },
+  tags: ["autodocs"],
+  argTypes: {
+    direction: { control: "radio", options: ["up", "down"] },
+    sizing: { control: "radio", options: ["longest", "current"] },
+    intervalMs: { control: { type: "number", min: 200, step: 100 } },
+    transitionMs: { control: { type: "number", min: 0, step: 20 } },
+    paused: { control: "boolean" },
+  },
+} satisfies Meta<typeof CyclingText>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+const STAGES = [
+  "Thinking",
+  "Reading messages",
+  "Drafting reply",
+  "Connecting the dots",
+  "Almost there",
+];
+
+export const Default: Story = {
+  args: {
+    items: STAGES,
+    intervalMs: 1800,
+    transitionMs: 380,
+    direction: "up",
+    sizing: "longest",
+  },
+};
+
+export const InsideAButton: Story = {
+  render: (args) => (
+    <Button variant="secondary">
+      <SpinnerIcon className="mr-2 size-4 animate-spin" />
+      <CyclingText {...args} />
+    </Button>
+  ),
+  args: {
+    items: STAGES,
+  },
+};
+
+export const InsideAStatusRow: Story = {
+  render: (args) => (
+    <div className="text-content-tertiary inline-flex items-center gap-2">
+      <SpinnerIcon className="size-4 animate-spin" />
+      <CyclingText {...args} />
+    </div>
+  ),
+  args: {
+    items: STAGES,
+  },
+};
+
+export const SizedToCurrent: Story = {
+  args: {
+    items: STAGES,
+    sizing: "current",
+  },
+};
+
+export const ShortItems: Story = {
+  args: {
+    items: ["One", "Two", "Three"],
+    intervalMs: 900,
+  },
+};
+
+export const Paused: Story = {
+  args: {
+    items: STAGES,
+    paused: true,
+  },
+};
+
+export const FakePlaceholder: Story = {
+  render: (args) => {
+    const [value, setValue] = useState("");
+    const [focused, setFocused] = useState(false);
+    const showPlaceholder = !focused && value.length === 0;
+
+    return (
+      <div className="relative w-80">
+        <input
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+          onFocus={() => setFocused(true)}
+          onBlur={() => setFocused(false)}
+          className="border-border-default bg-bg-primary text-content-primary w-full rounded-md border px-3 py-2 outline-none"
+        />
+        {showPlaceholder && (
+          <CyclingText
+            {...args}
+            className="text-content-tertiary pointer-events-none absolute inset-y-0 left-3 flex items-center"
+          />
+        )}
+      </div>
+    );
+  },
+  args: {
+    items: [
+      "Search creators…",
+      "Find a fan…",
+      "Look up a transaction…",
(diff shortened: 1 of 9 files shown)


SamFanvue:
Just a small thing. Ask your bot to add a demo to App.tsx

Gives us an easy area to mess around with components

simon-fanvue:
@SamFanvue @simonstaton If this looks okay to you, I'll put it in. Still needs micro-adjustments for timing but I'd like to get back to that once it’s in place in eden.

SamFanvue (approved):

src/components/CyclingText/CyclingText.stories.tsx
```diff
@@ -0,0 +1,116 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import { useState } from "react";
+import { Button } from "../Button/Button";
+import { SpinnerIcon } from "../Icons/SpinnerIcon";
+import { CyclingText } from "./CyclingText";
+
+const meta = {
+  title: "Components/CyclingText",
```
simonstaton, line 7: Please can you include a link to the Figma component in here

simonstaton (changes requested):

src/components/CyclingText/CyclingText.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    return (
+      <div className="relative w-80">
+        <input
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+          onFocus={() => setFocused(true)}
+          onBlur={() => setFocused(false)}
+          className="w-full rounded-md border border-border-default bg-bg-primary px-3 py-2 text-content-primary outline-none"
+        />
```
simonstaton: Do we need the input in the story for this component?

src/components/CyclingText/CyclingText.test.tsx
```diff
@@ -0,0 +1,175 @@
+import { act, render, screen } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { axe } from "vitest-axe";
+import { CyclingText } from "./CyclingText";
+
+const ITEMS = ["Alpha", "Beta", "Gamma"];
+
+const installMatchMedia = (matches: boolean) => {
+  Object.defineProperty(window, "matchMedia", {
```
simonstaton: Is there a safer/more global approach here?

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+
+const DEFAULT_INTERVAL_MS = 2100;
+const DEFAULT_TRANSITION_MS = 200;
+
+const SLIDE_OFFSET = "1.15em";
```
simonstaton: Can we avoid using em units for this, prefer px if not font scaling

src/components/CyclingText/CyclingText.stories.tsx
simon-fanvue: Done

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+    }, []);
+
+    if (itemCount === 0) {
+      return null;
+    }
+
+    const incomingOpacityEase = "cubic-bezier(0.72, 0, 0.92, 1)";
+    const incomingTransformEase = "cubic-bezier(0.42, 0, 0.58, 1)";
+    const outgoingOpacityEase = "cubic-bezier(0.42, 0, 1, 1)";
+    const outgoingTransformEase = "cubic-bezier(0.22, 1, 0.36, 1)";
```
simonstaton: Can we move this into tailwind utility classes?

simonstaton: Thinking about it I think a lot of this logic could be moved into tailwind for better standardisation:

Drive a `data-state="enter|exit"` attribute on each layer, then use `motion-safe:transition-[opacity,transform]           motion-safe:duration-(--cycling-duration) data-[state=exit]:opacity-0 data-[state=exit]:translate-y-(--cycling-offset-out) ...`, with a single `style={{ "--cycling-duration": ..., "--cycling-offset-out": ... }}` on the wrapper.

That gives you a single CSS-variable-bridging style prop and shifts the rest into Tailwind, mirroring how Tabs/TabsList.tsx:119 uses motion-safe:transition-[transform,width,height] motion-safe:duration-200 etc

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+   */
+  sizing?: CyclingTextSizing;
+  /**
+   * Class applied to each visible label span (current + incoming). Use this for
+   * effects that have to sit on the text element itself, e.g. `background-clip: text`.
+   */
+  labelClassName?: string;
+}
+
+const useReducedMotion = () => {
```
simonstaton: Lets avoid moving this into JS runtime and rely on native CSS/Tailwind

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+    const wrapperStyle: React.CSSProperties =
+      trackWidth !== null ? { width: `${trackWidth}px` } : {};
+
+    return (
+      <span
+        ref={ref}
+        data-testid="cycling-text"
+        data-paused={paused ? "true" : undefined}
+        className={cn(
+          "relative inline-block overflow-hidden transition-[width] duration-300 ease-out",
```
simonstaton: Needs motion safe selector

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+          aria-hidden="true"
+          className="pointer-events-none invisible inline-block select-none whitespace-nowrap"
+        >
+          {sizingLabel}
+        </span>
+
+        {/* Current (visible) item — slides up and fades out when transitioning. */}
+        <span
+          aria-live="polite"
+          aria-atomic="true"
```
simonstaton: Might want to check this with a screen-reader, I believe every time it cycles the text it will announce to the user which could be pretty annoying

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+    setReduced(mq.matches);
+    const onChange = (e: MediaQueryListEvent) => setReduced(e.matches);
+    mq.addEventListener("change", onChange);
+    return () => mq.removeEventListener("change", onChange);
+  }, []);
+
+  return reduced;
+};
+
+const useDocumentVisible = () => {
```
simonstaton: This feels like a bit of a hack, what is this trying to do?

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+      transitionMs = DEFAULT_TRANSITION_MS,
+      direction = "up",
+      paused = false,
+      sizing = "longest",
+      className,
+      labelClassName,
+      ...rest
+    },
+    ref,
+    // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: coordinates timers, resize observation, and motion layers
```
simonstaton: Lets avoid ignoring these if possible, normally this means the component should be broken into hooks or split out to be more granular

src/components/CyclingText/CyclingText.tsx
```diff
@@ -0,0 +1,353 @@
...
+  ) => {
+    const reducedMotion = useReducedMotion();
+    const docVisible = useDocumentVisible();
+
+    const [currentIndex, setCurrentIndex] = React.useState(0);
+    const [incomingIndex, setIncomingIndex] = React.useState<number | null>(null);
+    const [incomingEntered, setIncomingEntered] = React.useState(false);
+    const [transitioning, setTransitioning] = React.useState(false);
+
+    const currentIndexRef = React.useRef(0);
```
simonstaton: How come we can't use currentIndex for this?

simonstaton (changes requested):

simon-fanvue:
@simonstaton thanks, I'll work through these

src/components/CyclingText/CyclingText.stories.tsx
simon-fanvue: Removed the input

src/components/CyclingText/CyclingText.tsx
simon-fanvue: Switched to px

src/components/CyclingText/CyclingText.tsx
simon-fanvue: announceChanges false now

src/components/CyclingText/CyclingText.tsx
simon-fanvue: Moved to a use. To detect when the user tabs away to save animation cycles

src/components/CyclingText/CyclingText.tsx
simon-fanvue: Tried this way but I can't get the control on the animation to work out. Tailwind translate-y + CSS variables didn't reliably emit interpolable transforms

src/components/CyclingText/CyclingText.tsx
simon-fanvue: Also tried this one but  inline transition strings don't appear to cooperate with Tailwind without !important. Open to suggestions...

src/components/CyclingText/CyclingText.tsx
simon-fanvue: Switched to uses and removed the ignore

simonstaton (approved):

PR #469, opened 2026-05-11
feat(InlineEdit): add component and dotted Chip variant

morrid0:
- Adds `InlineEdit`: a chip-styled inline editor that swaps between a static label and a text input on click. Enter and blur commit the draft via `onCommit`; Escape reverts and calls `onCancel`. Optional `leftIcon` shows in display mode only (icon space is reclaimed in edit mode).
- Adds a `dotted` variant to `Chip` for dashed-border add/create affordances, with matching hover/active states.
- Stories, tests (axe + interaction + API), exports in `src/index.ts`, and an `App.tsx` demo entry.

Figma: `Fanvue-Library` nodes [17354-271](https://www.figma.com/design/S8zFdcOjt4qN4PrwntuCdt/Fanvue-Library?node-id=17354-271) (default) / [17354-265](https://www.figma.com/design/S8zFdcOjt4qN4PrwntuCdt/Fanvue-Library?node-id=17354-265) (hover).

### DIFF
--- a/src/components/InlineEdit/InlineEdit.tsx
+++ b/src/components/InlineEdit/InlineEdit.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import { cn } from "../../utils/cn";
+import { Chip } from "../Chip/Chip";
 
 /** Height of the inline edit field in pixels. */
 export type InlineEditSize = "32" | "40";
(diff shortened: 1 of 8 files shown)


src/components/InlineEdit/InlineEdit.tsx
```diff
@@ -0,0 +1,201 @@
...
+        cancel();
+      }
+    };
+
+    const showLeftIcon = Boolean(leftIcon) && !isEditing;
+    const chipClassName = cn(
+      "typography-semibold-body-sm inline-flex items-center rounded-xs border border-dashed border-border-primary bg-transparent px-3 text-content-primary motion-safe:transition-colors motion-safe:duration-150",
+      SIZE_CLASSES[size],
+      disabled && "pointer-events-none opacity-50",
+    );
```
simonstaton: This looks to reimplement the Chip's dotted styling instead of composing `<Chip dotted variant="square">`, is it not possible to use a composable Chip component here so we don't have to maintain in multiple places?

src/components/InlineEdit/InlineEdit.tsx
```diff
@@ -0,0 +1,201 @@
...
+              "absolute inset-0 block h-auto w-full min-w-0 outline-none focus-visible:shadow-focus-ring",
+            )}
+            {...inputProps}
+          />
+        ) : (
+          <button
+            type="button"
+            onClick={enterEditMode}
+            disabled={disabled}
+            aria-label={editLabel}
```
simonstaton: I don't think we need an aria-label for this, it should probably be the folder name `<span className="min-w-0 truncate">{value}</span>`

src/components/InlineEdit/InlineEdit.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+export interface InlineEditProps
+  extends Omit<
+    React.InputHTMLAttributes<HTMLInputElement>,
+    "value" | "defaultValue" | "size" | "onChange" | "onBlur" | "onKeyDown"
+  > {
+  /** Current value displayed in the chip and used as the starting draft when editing. */
+  value: string;
+  /** Called with the trimmed draft when the user commits an edit (Enter or blur). */
+  onCommit: (value: string) => void;
```
simonstaton: Nit: can we rename this onSubmit?

src/components/InlineEdit/InlineEdit.tsx
```diff
@@ -0,0 +1,201 @@
...
+      placeholder,
+      ...inputProps
+    },
+    ref,
+  ) => {
+    const [isEditing, setIsEditing] = React.useState(false);
+    const [draft, setDraft] = React.useState(value);
+    const inputRef = React.useRef<HTMLInputElement>(null);
+
+    React.useImperativeHandle(ref, () => inputRef.current as HTMLInputElement);
```
simonstaton: Is there a change inputRef.current is null and this ref is broken?

src/components/InlineEdit/InlineEdit.tsx
```diff
@@ -0,0 +1,201 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+
+/** Height of the inline edit field in pixels. */
+export type InlineEditSize = "32" | "40";
+
+export interface InlineEditProps
+  extends Omit<
+    React.InputHTMLAttributes<HTMLInputElement>,
+    "value" | "defaultValue" | "size" | "onChange" | "onBlur" | "onKeyDown"
```
simonstaton: Are we sure we don't want to bind to onChange etc? As this is an OSS package we should let consumers retain control and validation

src/components/InlineEdit/InlineEdit.tsx
```diff
@@ -0,0 +1,201 @@
...
+      input.select();
+    }, [isEditing]);
+
+    const exitEditMode = () => {
+      setIsEditing(false);
+    };
+
+    const commit = () => {
+      const trimmed = draft.trim();
+      if (trimmed.length === 0 || trimmed === value) {
```
simonstaton: Should we allow consumers to define their prop handlers via props incase they need to set their own validation rules?

simonstaton (changes requested):

simonstaton (approved):

PR #448, opened 2026-04-28
feat(ButtonV2): v2 Button with explicit negative variant (ENG-9997)

SamFanvue:
## ENG-9997

Adds `ButtonV2` (exported as `ButtonV2` / types `ButtonV2Props`, etc.) with negative-surface variants, token updates in `styleTokens.json` / `theme.css`, and `buildStyles.js` shadow handling for multi-layer AI button glow.

### Also
- `.local/` gitignored for local-only notes; removed `TOKENS_ADDED.md` from the component folder.

Made with [Cursor](https://cursor.com)

PR #480, opened 2026-05-15
[ui] ENG-9997 Evolve Button to v2 design tokens

alfonsovo-fv:
## ENG-9997

Picks up [#448](https://github.com/fanvue/fanv-ui/pull/448) by evolving `Button` in place instead of forking `ButtonV2`. Single source of truth, no `*V2` API.

## Why not `ButtonV2`?

The renames in #448 (`brand`→`upsell`, `destructive`→`error`, `white`→`alwaysWhite`) and removals (size `24`, `discount`/`price`, `link`/`tertiaryDestructive`/`text`) are not technically required by the new tokens. Figma confirms:

- `Brand`/`Upsell` use the same tokens (`--color-buttons-brand*`).
- `Destructive`/`Error` differ only in palette (red-700 → red-550/650).
- The "Buttons - Subscribe" frame still ships `discount`/`price` layouts in v2.
- All four sizes (24/32/40/48) are present in the v2 frame.
- `Link` was promoted to its own future component (separate "V2 Link" frame, sizes 14/16) — until we ship it, we shouldn't yank the variant.

## Summary

- **New `negative` boolean prop** for `primary`/`secondary`/`tertiary`/`outline`. Forces the dark-surface treatment regardless of theme. Other variants ignore it.
- **New variants**: `outline`, `ai`, `alwaysBlack`. The `ai` variant ships with the AI sparkle icon baked in (locked at 16px per Figma) and accepts `leftIcon` as an override.
- **Visual evolution under the same names**:
  - `secondary` was outlined → now filled `blackalpha-100`.
  - `tertiary` hover swaps from `brand-primary-muted` → `blackalpha-200`.
  - `destructive` swaps `red-700` → `red-550`/`red-650` (matches v2 `error`).
  - Disabled state now uses per-variant `bg-buttons-disabled-{default,negative}` and `text-content-disabled` instead of a blanket `opacity-50`.
- **Preserved (no breaking changes)**: `link`, `tertiaryDestructive`, `text`, size `24`, `discount`, `price`, `fullWidth`, `loading`, `asChild`, `leftIcon`/`rightIcon`.
- **Cherry-picked from #448**: `styleTokens.json` additions (`buttons.*` v2 tokens, `content.{disabled,alwaysWhite,alwaysBlack}`, `red.550`/`red.650` primitives, three-layer `aiButtonGlow` shadow), `buildStyles.js` N-layer drop-shadow loop, regenerated `theme.css`.
- Updated `App.tsx` showcase, stories (new variant + four `*Negative` stories + `AllStylesV2` matrix), and tests (a11y on new variants and on `negative`, parametric no-op assertion on non-negative-aware variants).

## Test plan

- [x] `pnpm typecheck` — 209 pre-existing errors on `main`, **zero new** (all in Drawer/DropdownMenu/InfoBox).
- [x] `pnpm lint:fix` — clean.
- [x] `pnpm test` — Button: 34/34 ✓; full unit suite: 2349/2349 ✓. Story tests rely on Playwright browser binary (CI handles).
- [ ] Local Storybook QA on `Components → Button → AllStylesV2` and the `*Negative` stories.
- [ ] Visual review: secondary/tertiary/destructive evolution, AI sparkle baked in, negative-on-dark.
- [ ] Confirm no Pandora/fadmin call sites regress (no API removals, all v1 names still work).

## Closes

Closes #448 in favour of this PR.

Made with Cursor

Made with [Cursor](https://cursor.com)


src/components/Button/Button.tsx
```diff
@@ -224,8 +306,11 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     ref,
   ) => {
     const Comp = asChild ? Slot : "button";
-    const isDisabled = disabled || loading;
-    const iconSizeClass = ICON_WRAPPER_CLASS[size];
+    const isDisabled = Boolean(disabled) || loading;
+    // AI variant locks its sparkle icon at 16px regardless of button size, per Figma.
+    const iconSizeClass = variant === "ai" ? "[&>svg]:size-4" : ICON_WRAPPER_CLASS[size];
+    // AI variant ships with the AI sparkle as a default left icon; consumer can override.
```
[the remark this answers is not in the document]
alfonsovo-fv: Good catch — fixed in 4710bad. Removed both inline comments and surfaced the AI default-icon behaviour in the component JSDoc instead so it shows up on consumer hover.

src/components/Button/Button.tsx
```diff
@@ -61,24 +80,85 @@ const ICON_WRAPPER_CLASS: Record<ButtonSize, string> = {
...
+        AI_GRADIENT_HOVER,
+      );
+    case "link":
+      return "bg-transparent text-content-primary underline decoration-solid hover:bg-buttons-tertiary-hover active:bg-buttons-tertiary-hover";
+    case "tertiaryDestructive":
+      return "bg-transparent text-error-content hover:bg-error-surface active:bg-error-surface";
+    case "text":
+      return "bg-transparent text-content-primary hover:underline active:underline";
+  }
+}
```
simonstaton: Anyway we can make this more readable? Can see this becoming hard to maintain compared to the 1 dimensional object from before

simonstaton (changes requested):

src/components/Button/Button.tsx
alfonsovo-fv: Good call — refactored in bc27fbe to bring back the flat lookup. There's now a single `VARIANT_CLASSES: Record<ButtonVariant, VariantClasses>` map with one row per variant; each row holds the four class strings (`default`, `disabled`, and optionally `negative` / `negativeDisabled`). The resolver collapses to:

```ts
const spec = VARIANT_CLASSES[variant];
const isNegative = NEGATIVE_AWARE_VARIANTS.has(variant) && negative;
if (disabled) return (isNegative && spec.negativeDisabled) || spec.disabled;
return (isNegative && spec.negative) || spec.default;
```

Adding a new variant is now a single object literal again, and the negative/disabled variations sit side-by-side per variant so they're easy to scan and diff.

simonstaton (approved):

PR #516, opened 2026-06-14
fix(ChartCard): prevent trendChip label truncation on mobile

simonstaton:
## Summary

- Adds `flex-wrap` to the subtitle container div so the comparison badge wraps to the next line on narrow screens instead of being clipped
- Adds `whitespace-nowrap` to the trendChip span so badge text (e.g. `$455.68 vs Mar`) never breaks mid-word inside the pill

Fixes https://linear.app/fanvue/issue/ENG-10098

## Test plan

- [ ] Verify `$455.68 vs Mar` badge renders fully on narrow mobile viewports in the earnings ChartCard
- [ ] New unit test: `renders trendChip label in full without truncation` passes
- [ ] All existing Chart tests continue to pass

### DIFF
--- a/src/components/Chart/Chart.test.tsx
+++ b/src/components/Chart/Chart.test.tsx
@@ -250,6 +250,22 @@ describe("Chart", () => {
     });
   });
 
+  describe("ChartCard", () => {
+    it("renders trendChip label in full without truncation", () => {
+      render(
+        <ChartCard
+          title="This month"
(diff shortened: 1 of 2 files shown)


simonstaton:
Closing as duplicate — PR #517 covers the same fix and has already been marked ready and enrolled for red-team review (workflow d20c9fa4).

PR #537, opened 2026-06-30
feat(icons): add UserItem, UserDisplayNameContainer, ProfileOnlineStatus, and VerifiedIcon

hronchetti:
## Summary

Ports the app-local `FVUserItem` into `@fanvue/ui` as a properly decomposed set of design-system components, decoupled from all app/Next/i18n/DB dependencies. The original was a single file depending on `@pandora/*`, `@eden/*`, `next/link`, `react-i18next` and a dark-mode hook; this PR re-expresses it in the repo's conventions (per-directory components, `forwardRef` + `displayName`, JSDoc, stories, tests, token-driven styling) with **no app coupling**.

Decisions taken with the maintainer along the way: dropped the `FV` prefix (DS naming), did **not** port `FVUserItemLink` (keeps the DS free of a routing/Next dependency), replaced `react-i18next` with English-defaulted label props, made `getInitials` dependency-free, and consolidated the verified/ambassador artwork onto a single `currentColor` `VerifiedIcon`.

## Changes

**New components** (each in its own directory with `.tsx` + `.stories.tsx` + `.test.tsx`):
- `UserItem` — avatar + display-name/nickname + handle row, with muted and online indicators. Accepts arbitrary pixel `avatarSize` and snaps to the nearest supported `Avatar` token. Root props tightened to `HTMLAttributes<HTMLDivElement>` (no `[key: string]: any`).
- `UserDisplayName` — display name with optional ambassador/verified badges and inline online status. i18n replaced with `ambassadorLabel` / `verifiedLabel` / `onlineLabel` props (English defaults); badges use `role="img"` + `aria-label`.
- `UserHandle` — `@`-prefixed, muted, truncated handle text.
- `ProfileOnlineStatus` — pulsing-dot + label online indicator, built on the existing `OnlineBlinkingIcon`.

**`UserDisplayName` API tightened (post-review):**
- Renamed `UserDisplayNameContainer` → `UserDisplayName` and `UserHandleTypography` → `UserHandle` (dropped the MUI-era suffixes now that these are first-class DS components).
- Removed the MUI-carryover layout/style props (`mt`, `pt`, `maxWidth`, `textAlign`, `color="white"`) — consumers express these with `className`.
- Removed the `variant` prop and its typography scale map. Every variant except the default mapped 1:1 to a `typography-*` token, and the odd ones out (`subtitle1`, `captionRegular`) were dropped. Typography is now supplied via `className` on the root and inherited by the name span, so the component no longer bakes in a scale (`UserItem` and the stories pass `typography-body-small-14px-semibold` explicitly). Root type is now `HTMLAttributes<HTMLElement>`.
- Ambassador badge takes precedence over verified when both are set. Its tint uses the theme-aware `text-success-content` (**not** the fixed `text-icons-brand-green`) so the green darkens in light mode and lightens in dark mode, matching `ProfileOnlineStatus`; the verified badge uses `text-content-primary`. Rationale lives in the component JSDoc (no inline comments in component/story/test files, per `AGENTS.md`).

**Truncation fix (`UserDisplayName`):**
- Fixed a bug where, with a long name, the trailing badge/online-status icons were clipped/lost instead of the name truncating. Previously the name text **and** the icons shared one element carrying `truncate` (`overflow:hidden; text-overflow:ellipsis; white-space:nowrap`), so the icons — being the last inline content — were pushed past the clip edge and cut off, and the ellipsis never landed on the name. (There is no CSS way to keep trailing inline icons visible while ellipsizing the same element's text, so this needed a structural change.)
- Restructured into a flex row (`inline-flex max-w-full items-center`) where only the name shrinks/truncates: the name lives in its own `min-w-0 … truncate` span, while the badge and `ProfileOnlineStatus` carry `shrink-0` and stay fully visible on the same line — nothing wraps or gets clipped.
- Removed the now-unnecessary vertical-centering hack on the badge (`-top-[25%]` / `translate-y-[25%]` / `h-full`); `items-center` handles alignment.
- Stories — **Truncated with icons + online status** (verified+online, ambassador+online, verified-only, online-only) and **Truncated (narrow container)** at `w-40` — demonstrate the name truncating while icons stay pinned. Unit tests that asserted truncation/typography on the root were updated to target the inner name span.

**Icons:**
- Added `VerifiedIcon` as a `currentColor` icon that themes via the wrapper's text colour. It renders at 16 / 24 / 32 px with `outlined` and `filled` variants (defaults to `filled`) — each size has its own tuned geometry rather than one path scaled down. The display-name badge passes `size={16}` to use the dedicated 16 px paths.
- The ambassador and verified badges both reuse `VerifiedIcon`, differing only by the tint token described above.
- Registered `VerifiedIcon` in `Icons.test.tsx`. (An earlier interim `AmbassadorIcon` was added then removed in favour of the shared icon.)

**Utils:**
- Added `getInitials` (+ tests), dependency-free (inlined whitespace cleanup; no `deverything` dependency).

**Barrel:**
- Exported `UserItem`, `UserDisplayName`, `UserHandle`, `ProfileOnlineStatus`, `VerifiedIcon`, and `getInitials` from `src/index.ts`.

## Checklist

- [x] TypeScript compiles (`pnpm typecheck`)
- [x] Linter passes (`pnpm lint`) — all changed files are clean; the only `pnpm lint` failures are pre-existing complexity warnings in `src/styles/buildStyles.js`, which this branch does not touch.
- [x] Tests pass (`pnpm test`) — full suite green (139 files, 3802 tests)
- [x] New/updated components have stories
- [x] New/updated components have accessibility tests
- [x] New/updated components have forwardRef unit tests
- [x] New/updated components have className chaining unit tests
- [x] Exported in `src/index.ts` (if consumable by vendors)
- [ ] Figma link added to story `design` parameter (if applicable) — **pending**: no Figma references attached yet to the new stories.
- [x] Does not have barrel file `src/YourComponent/index.ts`

### UserDisplayName
Screenshot 2026-06-30 at 13 51 35
Screenshot 2026-06-30 at 13 51 41

### ProfileOnlineStatus
Screenshot 2026-06-30 at 13 54 12
Screenshot 2026-06-30 at 13 52 38

### UserHandle
Screenshot 2026-06-30 at 13 54 06
Screenshot 2026-06-30 at 13 55 04

### UserItem
Screenshot 2026-06-30 at 13 55 53
Screenshot 2026-06-30 at 13 56 05


alfonsovo-fv (commented):
A few notes from a review, mostly about public API surface and a couple of convention slips. Nothing blocking.

src/components/UserDisplayNameContainer/UserDisplayNameContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /** Accessible label for the ambassador badge. @default "Ambassador" */
+  ambassadorLabel?: string;
+  /** Text colour override for rendering on dark surfaces. */
+  color?: "white";
+  /** HTML element used for the root wrapper. @default "span" */
+  component?: "span" | "h1" | "h2" | "h3" | "p" | "div";
+  /** Maximum width of the root wrapper. */
+  maxWidth?: string | number;
+  /** Top margin as a spacing multiplier, where `1` equals 8px. */
+  mt?: number;
```
alfonsovo-fv: These mt/pt/maxWidth/textAlign/color props feel like MUI-style leftovers from the app version rather than something we want in the DS public API. Consumers can already do all of this with className. I wonder if we drop them and keep the surface to just the badge/online/variant stuff? color="white" in particular is very narrow.

src/components/UserDisplayNameContainer/UserDisplayNameContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+    },
+    ref,
+  ) => {
+    const Component = component as React.ElementType;
+
+    const typographyClass =
+      variantClassMap[variant ?? "body2SemiBold"] ?? "typography-body-small-14px-semibold";
+
+    // `ambassador` takes precedence over `verified` when both are set.
+    // The ambassador tint uses `text-success-content` (not `text-icons-brand-green`)
```
alfonsovo-fv: AGENTS.md says no inline comments in component/story/test files. Could we move this tint rationale (and the precedence note) into the JSDoc, and drop the inline ones here, on the name span below, and in the stories/tests?

src/components/UserDisplayNameContainer/UserDisplayNameContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+      >
+        {/* Only the name shrinks/truncates; badges and status stay visible on the same line. */}
+        <span className={cn("min-w-0", typographyClass, noWrap && "truncate")}>{children}</span>
+        {badge && (
+          <span
+            role="img"
+            aria-label={badge.label}
+            className={cn("ml-2 inline-flex shrink-0 items-center", badge.tint)}
+          >
+            <VerifiedIcon className="size-4" />
```
alfonsovo-fv: Since size isn't passed this uses the 24px geometry scaled down to 16px via size-4. Could we pass size={16} so it uses the dedicated 16px paths?

src/components/UserDisplayNameContainer/UserDisplayNameContainer.tsx
```diff
@@ -0,0 +1,141 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+import { VerifiedIcon } from "../Icons/VerifiedIcon";
+import { ProfileOnlineStatus } from "../ProfileOnlineStatus/ProfileOnlineStatus";
+
+const variantClassMap: Record<string, string> = {
+  body2SemiBold: "typography-body-small-14px-semibold",
+  body1SemiBold: "typography-body-default-16px-semibold",
+  subtitle1: "text-lg font-semibold leading-[150%]",
```
alfonsovo-fv: Every other variant maps to a typography-* token but this one is raw utilities. Is there a token we should use, or should we drop subtitle1 if there's no matching scale?

src/components/UserItem/UserItem.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+      render(<UserItem user={baseUser} showAvatar={false} />);
+      expect(screen.queryByTestId("avatar")).not.toBeInTheDocument();
+    });
+
+    it("forwards the avatar src and alt from the user", () => {
+      render(<UserItem user={{ ...baseUser, avatarUri: { url: "/jane.jpg" } }} />);
+      const image = document.querySelector("img");
+      // Radix renders the image lazily, but the alt is taken from displayName.
+      expect(screen.getByTestId("avatar")).toBeInTheDocument();
+      if (image) {
```
alfonsovo-fv: This one can pass without asserting anything since the expect is inside if (image). If Radix doesn't render the img synchronously the test is a no-op. Could we findByRole("img") and assert the alt unconditionally, or assert on what we actually control?

src/components/Icons/VerifiedIcon.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * Verified badge icon. Renders at sizes 16, 24, or 32 px with outlined and filled variants.
+ * Defaults to the filled variant.
+ *
+ * @example
+ * ```tsx
+ * <VerifiedIcon size={24} />
+ * ```
+ */
+export const VerifiedIcon = React.forwardRef<SVGSVGElement, VerifiedIconProps>(
+  ({ filled = true, ...props }, ref) => (
```
alfonsovo-fv, line 72: Every other prop-based icon defaults filled to false (BaseIconProps). Was defaulting this one to true intentional? Just want to make sure we're not surprising consumers who expect the outlined default across the icon set.

src/components/ProfileOnlineStatus/ProfileOnlineStatus.tsx
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Inline online-status indicator: a pulsing dot followed by a label. Used by
+ * {@link UserDisplayNameContainer} when `showOnlineStatus` is set.
+ */
+export const ProfileOnlineStatus = React.forwardRef<HTMLSpanElement, ProfileOnlineStatusProps>(
+  ({ className, label = "Online", ...props }, ref) => (
+    <span
+      ref={ref}
+      className={cn(
+        "typography-description-12px-regular ml-2 inline-flex items-center gap-1 text-success-content",
```
alfonsovo-fv: The ml-2 is baked into the root so the standalone stories render with a left margin. Would it be cleaner to let the consumer (UserDisplayNameContainer) own the spacing so the component is neutral on its own?

src/components/UserHandleTypography/UserHandleTypography.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * <UserHandleTypography>jane_doe</UserHandleTypography>
+ * ```
+ */
+export const UserHandleTypography = React.forwardRef<HTMLSpanElement, UserHandleTypographyProps>(
+  ({ children, className, ...props }, ref) => {
+    return (
+      <span
+        ref={ref}
+        className={cn(
+          "typography-body-small-14px-regular block max-w-full truncate text-content-secondary text-sm",
```
alfonsovo-fv: typography-body-small-14px-regular already sets 14px so the trailing text-sm looks redundant.

src/components/Icons/VerifiedIcon.tsx
hronchetti: @alfonsovo-fv good point on this, the change was intentional as we currently don't use outlined versions anywhere in pandora for the VerifiedIcon. I can do one of the following:

1. Default to outline for consistency and set filled to true in all use cases of the icon
2. Remove the outline option
3. Leave filled as the default

Which do you think is best?

alfonsovo-fv (approved):

PR #555, opened 2026-07-03
feat(IconButton): migrate to V2 type system (ENG-11780)

alfonsovo-fv:
## Summary

Migrates `IconButton` to the standardised **V2 Button Icon** type system while preserving the public API.

- `primary` / `secondary` / `tertiary` now use the shared button colour tokens and gain a `negative` dark-surface treatment (along with new `outline`).
- **Shape is size-driven, per the V2 Figma:** only the **`24`** size is squared (`rounded-xs`); `32` / `40` / `48` stay **circular** (`rounded-full`). (Earlier revisions squared every V2 size — corrected after design review.)
- New variants: **`outline`**, **`error`**, **`white`**, **`black`**.
- New **`negative`** prop (honoured on `primary` / `secondary` / `tertiary` / `outline`, mirroring `Button`).
- New **`48`** size (V2 spec); existing `24/32/40/52/72` retained. The `40` icon is now `16px` to match the spec (`48` keeps its `24px` icon).
- Bespoke legacy variants (`brand`, `contrast`, `messaging`, `navTray`, `tertiaryDestructive`, `stop`, `microphone`) are **unchanged and stay circular** for backward compatibility.

**No public API removals** — the migration is additive.

### Deferred (follow-ups)
- **`upsell`** variant: no `buttons-upsell-*` design tokens exist yet.
- **`ai`** variant: owned by ENG-10621 (AI Button Style).

## Visual evidence

V2 variants × sizes (24/32/40/48) + disabled — note the squared `24` and circular `32/40/48`:

V2 matrix

`negative` treatment on a dark surface (primary / secondary / tertiary / outline):

V2 negative

Disabled via an ancestor `<fieldset disabled>` (no `disabled` prop) — all variants dim correctly, addressing the Codex P2 note:

Fieldset disabled

## Changes
- `IconButton.tsx` — V2 token-based variant map, `negative` prop, size-driven shape (`24` squared / larger circular), `48` size, `40` icon at 16px, and CSS `disabled:` fallbacks so `<fieldset disabled>` is styled.
- `IconButton.stories.tsx` — V2 stories (matrix, negative, per-variant, fieldset-disabled), size `48`, V2 Figma link.
- `IconButton.test.tsx` — size-driven shape and disabled-fallback behaviour.
- `App.tsx` — V2 demo rows (variants, negative, size `48`).

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm test src/components/IconButton` + `Drawer` (43 passed)
- [x] `pnpm build`
- [ ] Design review against V2 Figma (size-driven shape / tokens)

Closes ENG-11780

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -5,14 +5,13 @@ import { Count, type CountSize } from "../Count/Count";
 /**
  * Visual style variant of the icon button.
  *
- * V2 types: `primary`, `secondary`, `tertiary`, `outline`, `error`, `white`,
- * `black`. `primary`, `secondary`, `tertiary` and `outline` honour the
- * {@link IconButtonProps.negative} prop. Their shape is size-driven — squared at
- * the `24` size, circular otherwise.
+ * `primary`, `secondary`, `tertiary`, `outline`, `error`, `white` and `black`
+ * use the shared button colour tokens and a size-driven shape (squared at the
+ * `24` size, circular otherwise). Of these, `primary`, `secondary`, `tertiary`
+ * and `outline` also honour the {@link IconButtonProps.negative} prop.
  *
- * Legacy types (circular at all sizes, retained for backward compatibility):
- * `brand`, `contrast`, `messaging`, `navTray`, `tertiaryDestructive`, `stop`,
- * `microphone`.
+ * `brand`, `contrast`, `messaging`, `navTray`, `tertiaryDestructive`, `stop` and
+ * `microphone` are bespoke variants that stay circular at every size.
  */
 export type IconButtonVariant =
   | "primary"
@@ -61,11 +60,11 @@ const countSizeMap: Record<IconButtonSize, CountSize> = {
 };
 
 /**
- * V2 type-system variants. Their shape is size-driven (see {@link IconButton}):
- * the `24` size is squared (`rounded-xs`), every larger size stays circular.
- * Legacy variants remain circular at all sizes.
+ * Variants whose corner radius is size-driven (see {@link IconButton}): the `24`
+ * size is squared (`rounded-xs`), every larger size stays circular. All other
+ * variants are circular at every size.
  */
-const V2_VARIANTS = new Set<IconButtonVariant>([
+const SIZE_DRIVEN_SHAPE_VARIANTS = new Set<IconButtonVariant>([
   "primary",
   "secondary",
   "tertiary",
@@ -87,6 +86,7 @@ const DISABLED_FILL = "disabled:bg-buttons-disabled-default disabled:text-conten
 const DISABLED_FILL_NEGATIVE =
   "disabled:bg-buttons-disabled-negative disabled:text-content-disabled";
 const DISABLED_TRANSPARENT = "disabled:text-content-disabled";
+const DISABLED_OPACITY = "disabled:opacity-50";
 
 type VariantClasses = {
   default: string;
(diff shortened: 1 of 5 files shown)


src/components/IconButton/IconButton.tsx
```diff
@@ -105,14 +227,12 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
...
-          "cursor-pointer rounded-full transition-all duration-150 ease-in-out disabled:cursor-default",
-          // Size variants
+          "cursor-pointer transition-all duration-150 ease-in-out disabled:cursor-default",
+          "focus-visible:shadow-focus-ring",
+          SQUARE_VARIANTS.has(variant) ? "rounded-sm" : "rounded-full",
           sizeVariants[size],
-          // Variant styles
-          iconButtonVariants[variant],
-          // Manual CSS overrides
+          getVariantClasses(variant, negative, disabled),
```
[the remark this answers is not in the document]
alfonsovo-fv: Addressed in 9a380eb. The disabled treatment for V2 variants is now expressed with literal CSS `disabled:` classes (e.g. `disabled:bg-buttons-disabled-default`, `disabled:text-content-disabled`, `disabled:border-buttons-disabled-*`) instead of switching purely on the `disabled` prop, and hover is guarded with `not-disabled:`. So a button disabled by an ancestor `<fieldset disabled>` now gets the disabled styling too, not just the prop path. Added a story (`FieldsetDisabled`) and a test asserting the fallback — see the new screenshot in the PR description.

src/components/IconButton/IconButton.tsx
```diff
@@ -2,67 +2,182 @@ import * as React from "react";
...
+  tertiaryDestructive:
+    "bg-transparent text-error-content hover:bg-brand-primary-muted not-disabled:active:bg-brand-primary-muted disabled:opacity-50",
+  stop: "bg-buttons-primary-default text-content-primary-inverted hover:bg-buttons-brand-default hover:text-content-always-black not-disabled:active:bg-buttons-brand-default not-disabled:active:text-content-always-black disabled:opacity-50",
+  microphone:
+    "bg-buttons-primary-default text-content-primary-inverted hover:bg-buttons-brand-default hover:text-content-always-black not-disabled:active:bg-buttons-brand-default not-disabled:active:text-content-always-black disabled:opacity-50",
+};
+
+function getVariantClasses(variant: IconButtonVariant, negative: boolean): string {
+  const v2 = V2_VARIANT_CLASSES[variant];
+  if (v2) {
```
simonstaton: Do we need this v2 branch? Can't we just release this to 100%?

simonstaton (changes requested):

src/components/IconButton/IconButton.tsx
alfonsovo-fv: Merged the two maps into a single `VARIANT_CLASSES` and dropped the `if (v2)` branch in `getVariantClasses` — no more V1/V2 split in the code (also renamed the leftover `V2_VARIANTS` set to `SIZE_DRIVEN_SHAPE_VARIANTS`, since it only drove the squared-at-24 shape rule).

Kept every variant in the one map rather than deleting them: `brand`/`contrast`/`messaging`/`navTray`/`tertiaryDestructive`/`stop`/`microphone` are still referenced internally (e.g. `Banner` uses `contrast`) and by external consumers, so this stays non-breaking rather than forcing a major. If design intends to fully retire those, we can add them to V2 in Figma or deprecate + migrate call sites in a follow-up.

Pushed in 7ff0c8a. Behaviour is identical (same class output); typecheck, biome, and the IconButton tests pass.

simonstaton (approved):

PR #567, opened 2026-07-08
feat(Badge): update to V2 design

morrid0:
Aligns the Badge component with the V2 Figma design, adds an alwaysWhite variant, and adds an all-variants Storybook story.


alfonsovo-fv (approved):

src/components/Badge/Badge.tsx
```diff
@@ -1,30 +1,31 @@
...
   variant: {
-    default: "bg-neutral-alphas-50 text-content-primary",
+    default: "bg-buttons-secondary-default text-content-primary",
     dark: "bg-neutral-alphas-600 text-content-always-white",
-    success: "bg-success-surface text-success-content",
-    warning: "bg-warning-surface text-warning-content",
-    error: "bg-error-surface text-error-content",
-    special: "bg-neutral-alphas-50 text-content-primary",
-    info: "bg-info-surface text-info-content",
+    success: "bg-buttons-secondary-default text-content-primary",
```
simonstaton, line 9: Are you sure this is right? This should be a semantic token for the badge like `bg-badge-error`

simonstaton (changes requested):

src/components/Badge/Badge.tsx
morrid0: Confirmed against the V2 Figma frame (node 18028-289). These plain status variants are intentionally neutral grey pills. The background binds to `Color/Buttons/Secondary/Default` in Figma (`#1515151a`), which is `bg-buttons-secondary-default` here, and only the status dot is coloured. There is no `Badge/*` colour token in the library. The tinted versions are the separate `*Colour` variants (`errorColour` -> `bg-error-surface`, etc.), which do use the semantic surface tokens.

morrid0: What about this one @simonstaton?

simonstaton (approved):

simonstaton (approved):

PR #585, opened 2026-07-09
feat(Inputs): add PhoneField and TextField action slot, align input hint styling to V2 board

morrid0:
Adds a PhoneField component and a trailing action slot on TextField for inline buttons, and aligns TextField/TextArea helper-text colour and spacing to the V2 Input Fields design.

### DIFF
--- a/src/components/PhoneField/PhoneField.tsx
+++ b/src/components/PhoneField/PhoneField.tsx
@@ -149,6 +149,25 @@ export const PhoneField = React.forwardRef<HTMLInputElement, PhoneFieldProps>(
         .filter(Boolean)
         .join(" ") || undefined;
 
+    const innerRef = React.useRef<HTMLInputElement>(null);
+    const setRefs = React.useCallback(
+      (node: HTMLInputElement | null) => {
+        innerRef.current = node;
+        if (typeof ref === "function") ref(node);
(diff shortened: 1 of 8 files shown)


src/components/PhoneField/PhoneField.tsx
```diff
@@ -0,0 +1,247 @@
...
+            className="typography-description-12px-semibold pb-2 text-content-primary"
+          >
+            {label}
+          </label>
+        )}
+
+        {/* biome-ignore lint/a11y/noStaticElementInteractions: focus bridge delegates pointer clicks on the padding to the input */}
+        <div
+          className={getContainerClassName(size, error, disabled)}
+          onMouseDown={handleContainerMouseDown}
```
simonstaton: Can't we just use onClick in the button and apply the styling there?

simonstaton (changes requested):

src/components/PhoneField/PhoneField.tsx
morrid0: Good shout. Dropped the focus-bridge handler and the a11y ignore, so the container div is now purely presentational and the button keeps its own onClick. Kept the styling on the div rather than the button: a button can't wrap the input, and a label wrapper would leak the dial code into the input's accessible name.

simonstaton (approved):

PR #588, opened 2026-07-10
feat(ReceiveMoneyIcon): add ReceiveMoneyIcon to iconography

morrid0:
Adds ReceiveMoneyIcon to the shared iconography with dedicated 16/24/32 geometry so the fan spendings icon renders consistently instead of being copied directly into the app.


simonstaton (approved):
Just checking, you generated this one with the icon script?

morrid0:
Yep, generated via `pnpm icons:sync` (extract from Figma + generate). All six changed files are the standard pipeline output — component, manifest/tags, and the auto-generated test/stories/barrel. No manual edits.

PR #600, opened 2026-07-20
fix(forms): keep focus ring visible on autofilled inputs

simonstaton:
## Summary

- On Chrome autofill, focused TextFields showed a broken focus ring — only the corner arcs remained, looking like a radius mismatch between the input and its outline.
- Root cause: the focus ring is an inset box-shadow on the field container, and inset shadows paint *below* child content. The autofill neutralisation hack in `base.css` puts an opaque `inset 0 0 0 1000px` shadow on the full-height `<input>`, which painted over the ring's top and bottom edges.
- Fix: render the ring in an `::after` overlay (`after:inset-0 after:rounded-[inherit]` + `has-focus-visible:after:shadow-focus-ring`) so nothing the input paints can erase it. Same token, same geometry. Applied to TextField, TextArea, PhoneField and Autocomplete; the paint-order gotcha is documented in `base.css`.

Autofilled inputs now look identical to manually-typed ones, with the focus ring intact.

## Test plan

- [x] Unit tests for TextField, TextArea, PhoneField, Autocomplete (271 passing)
- [x] `pnpm typecheck` and `pnpm lint`
- [x] Visual check in Storybook with simulated autofill shadow + forced `:focus-visible` (light + dark verified on local.fanvue.com signin)

### DIFF
--- a/src/styles/base.css
+++ b/src/styles/base.css
@@ -21,6 +21,13 @@
    * visibly lighter band inside the field in dark mode. Note this rule
    * cannot be overridden by unlayered !important declarations downstream:
    * for !important the cascade reverses, so @layer base wins.
+   *
+   * Side effect: this opaque shadow paints ABOVE any inset box-shadow on an
+   * ancestor (a child's background/shadow always paints over the parent's
+   * inset shadow), so field containers must not draw their focus ring as a
+   * plain inset shadow on themselves — it gets erased behind the autofilled
+   * input. TextField/TextArea/PhoneField/Autocomplete render the ring in an
+   * ::after overlay instead.
    */
   input:-webkit-autofill,
   input:-webkit-autofill:hover,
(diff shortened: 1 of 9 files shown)


src/styles/base.css
```diff
@@ -21,6 +21,13 @@
    * visibly lighter band inside the field in dark mode. Note this rule
    * cannot be overridden by unlayered !important declarations downstream:
    * for !important the cascade reverses, so @layer base wins.
+   *
+   * Side effect: this opaque shadow paints ABOVE any inset box-shadow on an
```
alfonsovo-fv, line 25: Nice writeup on the paint order. Worth a quick regression test somewhere that asserts the container renders the ring via the after overlay? I know autofill itself is basically untestable in jsdom, just thinking someone could revert the class string and nothing would catch it.

alfonsovo-fv (commented):
Nice fix, the root-cause writeup is spot on.

One scope question: InlineEdit does the same thing, its input sits inside a Chip that draws focus-within:shadow-focus-ring as an inset shadow on a positioned wrapper, and the DropdownMenu search field container is the same pattern too. Was leaving those out deliberate? Guessing they're just much less likely to ever get autofilled so happy either way, just want to confirm it was intentional.

alfonsovo-fv (approved):

PR #618, opened 2026-07-28
feat(CreatorTile): add banner ratio, and AI-disclosure badge to user components

jacksonford-07:
Additive changes so a creator tile can be rendered as a slim inline banner, and so a user's name can disclose that the account is AI-operated.

Everything here is an addition to an existing component. No removals, and no changes to how current props render — with one deliberate exception, called out under **Behaviour change** below.

## Changes

**`CreatorTile` — `aspectRatio="banner"` (5:2)**
The existing presets stop at 9:5, which is still too tall for a tile placed inline in a narrow column. `banner` sits at the short end of the scale.

Named for the slot rather than the axis: `wide` (its name in the first three commits) read on a different axis to `tall`/`medium`/`short`, so from autocomplete alone you could not tell whether it sat above or below `short`. Renamed before release, so no consumer is affected.

This commit also corrects the `aspectRatio` JSDoc, which documented all three existing ratios incorrectly: it claimed 1:2, 361:200 and 4:5 where the code has 5:4, 3:2 and 9:5. I would normally leave adjacent code alone, but adding a fourth entry next to three wrong ones would have compounded a mistake in a public API reference.

**`UserDisplayName` — `aiDisclosure` + `aiDisclosureLabel`**
Surfaces the existing `AIDisclosureIcon` beside the name so callers stop composing it by hand at each call site.

It renders _alongside_ the verified or ambassador badge rather than replacing it. Disclosure describes how an account is operated, so it is orthogonal to the status badges, which stay mutually exclusive with each other.

**`UserItem` — forwards `verified`, `aiDisclosure` and their labels**
`UserItem` already renders `UserDisplayName` but never passed the badge props through, so a row could not show either badge without rebuilding it by hand. The labels are forwarded too, so a non-English consumer can localise them; both stay `undefined` here so `UserDisplayName` remains the single source of the default strings.

## Behaviour change

**The verified and AI-disclosure badges now inherit the name's colour** (`text-current`) instead of hardcoding `text-content-primary`.

The hardcoded tint is theme-dependent, which breaks down on any surface that fixes its own text colour. `CreatorTile`'s `name` slot is exactly that surface: it sits over a cover image with white text, so in light mode the name and tagline stayed white while the badges flipped to black.

On a default surface this is a no-op, because the colour inherited there is the same `text-content-primary` that was hardcoded. What does change: **a consumer rendering the name in a muted or branded colour now gets a badge in that colour** rather than `text-content-primary`. That is the point of the fix, but it ships to consumers, so shout if you would rather it were opt-in behind a prop.

The ambassador badge deliberately keeps its own tint, since its green is mode-aware by design rather than context-dependent. There is a test pinning that, so the inheritance change cannot leak into it.

## Known limitation: minimum width

The profile row is a fixed ~74px tall, so at 5:2 the tile becomes shorter than its own row below **185px** of width and clips the top of the row. Measured in Storybook:

| Width  | Tile height | Row clipped |
| ------ | ----------- | ----------- |
| 160px  | 64px        | 10px        |
| 180px  | 72px        | 2px         |
| 185px  | 74px        | none        |
| 200px+ | 80px+       | none        |

Documented in the `aspectRatio` JSDoc rather than fixed, because the only thing that actually fixes it is an explicit pixel `min-height` — `min-height: fit-content` and `max-content` are both ignored, since `aspect-ratio` wins over the intrinsic contribution. That would hardcode the row's height into the container, which then rots the moment the row changes. Happy to add it if you would rather have the guard than the note.

Worth separating from a second effect at narrow widths: below ~200px the name and tagline collapse to zero width, squeezed out by the fixed-width action button. That one is **not** specific to this ratio — it reproduces identically on `tall`, `medium` and `short`, so it is pre-existing and out of scope here.

## Verification

- `pnpm typecheck` clean
- `pnpm lint` — 36 warnings, 0 errors, **identical to `main`** (verified by stashing and re-running, so this branch adds nothing)
- `pnpm test` — 3,487 unit tests and 953 Storybook browser tests pass
- `pnpm build` succeeds
- No e2e specs exist for these components
- Visually verified in Storybook, both themes: the 5:2 tile sits correctly at the short end of the ratio scale, badges render white over the banner in light and dark, and both badges keep the right order and spacing in `UserItem`

New stories cover each addition, so Chromatic should show the diffs.

### DIFF
--- a/src/components/CreatorTile/CreatorTile.tsx
+++ b/src/components/CreatorTile/CreatorTile.tsx
@@ -3,13 +3,13 @@ import { cn } from "../../utils/cn";
 import { Avatar } from "../Avatar/Avatar";
 
 /** Width-to-height ratio preset for the tile. */
-export type CreatorTileAspectRatio = "tall" | "medium" | "short" | "wide";
+export type CreatorTileAspectRatio = "tall" | "medium" | "short" | "banner";
 
 const ASPECT_RATIO_CLASSES: Record<CreatorTileAspectRatio, string> = {
   tall: "aspect-5/4",
   medium: "aspect-3/2",
   short: "aspect-9/5",
-  wide: "aspect-5/2",
+  banner: "aspect-5/2",
 };
 
 export interface CreatorTileProps extends React.HTMLAttributes<HTMLDivElement> {
@@ -32,7 +32,12 @@ export interface CreatorTileProps extends React.HTMLAttributes<HTMLDivElement> {
    * - `tall` – 5:4, closest to square
    * - `medium` – 3:2 landscape
    * - `short` – 9:5 landscape
-   * - `wide` – 5:2, a slim banner
+   * - `banner` – 5:2, the slimmest preset
+   *
+   * The profile row is a fixed ~74px tall, so `banner` needs at least 185px of
+   * width before the ratio makes the tile shorter than its own row and the top
+   * of the row is clipped. The taller presets reach that floor at widths too
+   * narrow to be practical.
    *
    * @default "medium"
    */
--- a/src/components/CreatorTile/CreatorTile.tsx
+++ b/src/components/CreatorTile/CreatorTile.tsx
@@ -3,12 +3,13 @@ import { cn } from "../../utils/cn";
 import { Avatar } from "../Avatar/Avatar";
 
 /** Width-to-height ratio preset for the tile. */
-export type CreatorTileAspectRatio = "tall" | "medium" | "short";
+export type CreatorTileAspectRatio = "tall" | "medium" | "short" | "wide";
 
 const ASPECT_RATIO_CLASSES: Record<CreatorTileAspectRatio, string> = {
   tall: "aspect-5/4",
   medium: "aspect-3/2",
   short: "aspect-9/5",
+  wide: "aspect-5/2",
 };
 
 export interface CreatorTileProps extends React.HTMLAttributes<HTMLDivElement> {
@@ -28,9 +29,10 @@ export interface CreatorTileProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
    * Width-to-height ratio preset.
    *
-   * - `tall` – 1:2 narrow portrait
-   * - `medium` – 361:200 landscape banner (default)
-   * - `short` – 4:5 closer to square
+   * - `tall` – 5:4, closest to square
+   * - `medium` – 3:2 landscape
+   * - `short` – 9:5 landscape
+   * - `wide` – 5:2, a slim banner
    *
    * @default "medium"
    */
--- a/src/components/UserDisplayName/UserDisplayName.test.tsx
+++ b/src/components/UserDisplayName/UserDisplayName.test.tsx
@@ -103,6 +103,40 @@ describe("UserDisplayName", () => {
       render(<UserDisplayName>Aitana</UserDisplayName>);
       expect(screen.queryByRole("img")).not.toBeInTheDocument();
     });
+
+    it("renders an AI-disclosure badge with an accessible label", () => {
+      render(<UserDisplayName aiDisclosure>Aitana</UserDisplayName>);
+      expect(screen.getByRole("img", { name: "AI creator" })).toBeInTheDocument();
+    });
+
+    it("supports a custom AI-disclosure label", () => {
+      render(
+        <UserDisplayName aiDisclosure aiDisclosureLabel="Operated with AI">
+          Aitana
+        </UserDisplayName>,
+      );
+      expect(screen.getByRole("img", { name: "Operated with AI" })).toBeInTheDocument();
+    });
+
+    it("renders the AI-disclosure badge alongside a status badge rather than replacing it", () => {
+      render(
+        <UserDisplayName aiDisclosure verified>
+          Aitana
+        </UserDisplayName>,
+      );
+      expect(screen.getByRole("img", { name: "Verified" })).toBeInTheDocument();
+      expect(screen.getByRole("img", { name: "AI creator" })).toBeInTheDocument();
+    });
+
+    it("renders the AI-disclosure badge alongside the ambassador badge", () => {
+      render(
+        <UserDisplayName aiDisclosure ambassador>
+          Aitana
+        </UserDisplayName>,
+      );
+      expect(screen.getByRole("img", { name: "Ambassador" })).toBeInTheDocument();
+      expect(screen.getByRole("img", { name: "AI creator" })).toBeInTheDocument();
+    });
   });
 
   describe("online status", () => {
@@ -130,6 +164,15 @@ describe("UserDisplayName", () => {
     it("has no accessibility violations with badges and online status", async () => {
       const { container } = render(
         <UserDisplayName ambassador showOnlineStatus>
+          Aitana Lopez
+        </UserDisplayName>,
+      );
+      expect(await axe(container)).toHaveNoViolations();
+    });
+
+    it("has no accessibility violations with the AI-disclosure badge", async () => {
+      const { container } = render(
+        <UserDisplayName aiDisclosure verified showOnlineStatus>
           Aitana Lopez
         </UserDisplayName>,
       );
--- a/src/components/UserDisplayName/UserDisplayName.test.tsx
+++ b/src/components/UserDisplayName/UserDisplayName.test.tsx
@@ -104,6 +104,24 @@ describe("UserDisplayName", () => {
       expect(screen.queryByRole("img")).not.toBeInTheDocument();
     });
 
+    // Surfaces that fix their own text colour, such as a row over a cover image, need the
+    // badges to follow it. A theme-dependent tint renders black against white text in light
+    // mode.
+    it("inherits the name's colour for the verified and AI-disclosure badges", () => {
+      render(
+        <UserDisplayName verified aiDisclosure>
+          Aitana
+        </UserDisplayName>,
+      );
+      expect(screen.getByRole("img", { name: "Verified" })).toHaveClass("text-current");
+      expect(screen.getByRole("img", { name: "AI creator" })).toHaveClass("text-current");
+    });
+
+    it("keeps the ambassador badge on its own tint rather than inheriting", () => {
+      render(<UserDisplayName ambassador>Aitana</UserDisplayName>);
+      expect(screen.getByRole("img", { name: "Ambassador" })).toHaveClass("text-success-content");
+    });
+
     it("renders an AI-disclosure badge with an accessible label", () => {
       render(<UserDisplayName aiDisclosure>Aitana</UserDisplayName>);
       expect(screen.getByRole("img", { name: "AI creator" })).toBeInTheDocument();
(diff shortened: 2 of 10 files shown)


jacksonford-07:
Going to update icons as they look off rn 
image

jacksonford-07:
Pushed a fourth commit that changes existing behaviour, so calling it out explicitly rather than leaving it in the review question above.

**`fix(UserDisplayName)`: the verified and AI-disclosure badges now inherit the name's colour.**

Both hardcoded `text-content-primary`, which is theme-dependent. On a surface that fixes its own text colour that breaks down, and `CreatorTile`'s `name` slot is exactly that surface: it sits over a cover image with white text, so in light mode the name and tagline stayed white while the badges flipped to black.

Verified visually in Storybook, both themes:

- **Wide tile, light** — badges were black beside white text, now white
- **Wide tile, dark** — unchanged, already white
- **Plain surface (`UserDisplayName`, `UserItem`), both themes** — unchanged, because the colour inherited there is the same `text-content-primary` that was hardcoded

The ambassador badge deliberately keeps its own tint, since its green is mode-aware by design rather than context-dependent. Added a test pinning that, so the inheritance change cannot leak into it.

The one case that does change: a consumer setting a non-default text colour on the root will now have the verified badge follow it. That is the intent of the fix, but it is a behaviour change on a public component, so shout if you would rather it were opt-in behind a prop and I will rework it.

Full suite after the change: `typecheck` clean, `lint` 36 warnings / 0 errors (identical to `main`), 3,485 unit and 953 Storybook tests passing.

jacksonford-07:
image

fixed

src/components/CreatorTile/CreatorTile.tsx
```diff
@@ -3,12 +3,13 @@ import { cn } from "../../utils/cn";
...
 
 /** Width-to-height ratio preset for the tile. */
-export type CreatorTileAspectRatio = "tall" | "medium" | "short";
+export type CreatorTileAspectRatio = "tall" | "medium" | "short" | "wide";
 
 const ASPECT_RATIO_CLASSES: Record<CreatorTileAspectRatio, string> = {
   tall: "aspect-5/4",
   medium: "aspect-3/2",
   short: "aspect-9/5",
+  wide: "aspect-5/2",
```
alfonsovo-fv: `CreatorTile.test.tsx` has an `it.each` over the ratios and it still only lists tall/medium/short, so `wide` ends up the one preset with nothing asserting it maps to `aspect-5/2`. Could you add the row? The `Record` keeps the key exhaustive but nothing pins the class itself.

src/components/CreatorTile/CreatorTile.tsx
```diff
@@ -3,12 +3,13 @@ import { cn } from "../../utils/cn";
 import { Avatar } from "../Avatar/Avatar";
 
 /** Width-to-height ratio preset for the tile. */
-export type CreatorTileAspectRatio = "tall" | "medium" | "short";
+export type CreatorTileAspectRatio = "tall" | "medium" | "short" | "wide";
```
alfonsovo-fv: `wide` reads on a different axis to `tall` / `medium` / `short`, so from autocomplete alone I can't tell whether it's shorter or taller than `short`. Would `shortest` or `banner` be clearer? Not blocking, just thinking about how the union reads to a consumer.

src/components/CreatorTile/CreatorTile.tsx
```diff
@@ -28,9 +29,10 @@ export interface CreatorTileProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
    * Width-to-height ratio preset.
    *
-   * - `tall` – 1:2 narrow portrait
-   * - `medium` – 361:200 landscape banner (default)
-   * - `short` – 4:5 closer to square
+   * - `tall` – 5:4, closest to square
+   * - `medium` – 3:2 landscape
+   * - `short` – 9:5 landscape
+   * - `wide` – 5:2, a slim banner
```
alfonsovo-fv: At 5:2 the height is width / 2.5, and the profile row needs roughly 72px (40px avatar plus `p-4`), so somewhere under ~180px wide I'd expect the row to start getting clipped at the top by the `overflow-hidden`. The stated use case is a narrow column, and every story is 280px or wider, so I couldn't tell from Chromatic. Did you check it at the narrow end?

src/components/UserDisplayName/UserDisplayName.tsx
```diff
@@ -63,7 +80,7 @@ export const UserDisplayName = React.forwardRef<HTMLElement, UserDisplayNameProp
     const badge = ambassador
       ? { label: ambassadorLabel, tint: "text-success-content" }
       : verified
-        ? { label: verifiedLabel, tint: "text-content-primary" }
+        ? { label: verifiedLabel, tint: "text-current" }
```
alfonsovo-fv, line 83: The PR body still says you left `verified` alone and asks whether we'd want `currentColor`, but the last commit does it. Could you update the description? I'd rather the next reviewer didn't skim past it.

I think the change is right, but it isn't quite a no-op: anyone rendering the name inside a muted or branded colour gets a muted tick now where they used to get `text-content-primary`. Fine by me, I just want it called out since it ships to consumers.

src/components/UserDisplayName/UserDisplayName.tsx
```diff
@@ -82,6 +99,15 @@ export const UserDisplayName = React.forwardRef<HTMLElement, UserDisplayNameProp
             <VerifiedIcon size={16} />
           </span>
         )}
+        {aiDisclosure && (
+          <span
+            role="img"
+            aria-label={aiDisclosureLabel}
+            className="ml-2 inline-flex shrink-0 items-center text-current"
```
alfonsovo-fv, line 106: The badge is labelled for screen readers only, so a sighted user just sees a glyph. Same as `verified` today, so I'm not asking you to solve it here, but disclosure feels like the one badge where a tooltip might actually be needed. Worth a follow-up?

src/components/UserDisplayName/UserDisplayName.test.tsx
```diff
@@ -103,6 +103,58 @@ describe("UserDisplayName", () => {
       render(<UserDisplayName>Aitana</UserDisplayName>);
       expect(screen.queryByRole("img")).not.toBeInTheDocument();
     });
+
+    // Surfaces that fix their own text colour, such as a row over a cover image, need the
```
alfonsovo-fv: AGENTS.md asks for no inline comments in component, story or test files. The rationale is already in the component JSDoc so I think we can just drop these three lines.

src/components/UserItem/UserItem.tsx
```diff
@@ -85,7 +91,11 @@ export const UserItem = React.forwardRef<HTMLDivElement, UserItemProps>(
           />
         )}
         <div className="flex-1 overflow-hidden pl-2">
-          <UserDisplayName className="typography-body-small-14px-semibold">
+          <UserDisplayName
+            aiDisclosure={aiDisclosure}
```
alfonsovo-fv, line 101: We forward the booleans but not `verifiedLabel` / `aiDisclosureLabel`, so a UserItem consumer is stuck with the English defaults in the a11y label with no way to override. Could we pass those through too?

alfonsovo-fv (commented):
Nice, this is clean and genuinely additive. A few things below: one missing test, a stale description, and a localisation gap. The rest are questions rather than blockers.

src/components/CreatorTile/CreatorTile.tsx
jacksonford-07: Added, in `d883db6`. Good catch that the `Record` only kept the key exhaustive — nothing pinned the class.

src/components/CreatorTile/CreatorTile.tsx
jacksonford-07: Gone with `banner`, in `d883db6`. Your framing convinced me: naming it for the slot sidesteps the axis problem entirely, where `shortest` would still leave a consumer guessing how it relates to `short`. Renamed everywhere including the showcase app, and the ratio never shipped so nobody is affected.

src/components/CreatorTile/CreatorTile.tsx
jacksonford-07: No, I had not — every story is 280px+ so Chromatic could not have shown me either. Measured it properly in Storybook, and your estimate was almost exact. The row is 74px, so clipping starts below **185px** (74 × 2.5):

| Width | Tile height | Row clipped |
|---|---|---|
| 160px | 64px | 10px |
| 180px | 72px | 2px |
| 185px | 74px | none |
| 200px+ | 80px+ | none |

Two things I found while measuring:

**The only thing that fixes it is an explicit pixel `min-height`.** `min-height: fit-content` and `max-content` are both *applied* and then ignored, because `aspect-ratio` wins over the intrinsic contribution. So the guard would have to be `min-h-[4.625rem]`, which hardcodes the row height into the container and rots the moment the row changes. I documented the floor in the JSDoc instead — say the word and I will add the guard, it is one class.

**There is a second, worse effect at narrow widths that is not this ratio.** Below ~200px the name and tagline collapse to zero width, squeezed out by the fixed-width action button, so you get an avatar and a button and no creator. That reproduces identically on `tall`, `medium` and `short` at the same width, so it is pre-existing and I left it alone — but it means the practical floor for a tile *with an action* is ~220px regardless of ratio, which is a stricter bound than the clipping one. Worth its own issue?

src/components/UserDisplayName/UserDisplayName.tsx
jacksonford-07: Description rewritten — the question is gone and the change now has its own **Behaviour change** section, with your muted/branded case stated explicitly as the thing that is not a no-op. Sorry, that was a stale body from before the fourth commit.

src/components/UserDisplayName/UserDisplayName.test.tsx
jacksonford-07: Dropped, in `819b98d`. Agreed, the JSDoc already carries it.

src/components/UserItem/UserItem.tsx
jacksonford-07: Done in `040a05e`, with a test. Both stay `undefined` rather than repeating the defaults, so `UserDisplayName` remains the only place the English strings live.

This one lands on us directly, so thanks for catching it — the pandora consumer needs `t()`-driven labels, and without the pass-through it would have shipped hardcoded English into every locale.

src/components/UserDisplayName/UserDisplayName.tsx
jacksonford-07: Not for now, and for a product reason rather than a scope one.

Today the AI disclosure is not introduced to a fan until they reach the profile. Adding an explanatory tooltip at this point in the funnel would surface it earlier than it currently appears, which risks hurting conversion on the very surface we are adding this for. The icon itself may already do the disclosure work without the label.

So: leaving it glyph-only, matching `verified`. If we later decide the disclosure should be explicit earlier, it is worth doing to both badges at once rather than just this one.

alfonsovo-fv (approved):

PR #620, opened 2026-07-30
chore(build): copy the token stylesheets from the build itself

jacksonford-07:
Makes the build emit `dist/styles/` itself, instead of relying on a trailing `cp` in the `build` npm script.

## Why

`vite build --watch` never ran the `cp`, so every watch rebuild left `dist/styles/` missing. That breaks `pnpm start:localui` in pandora: eden imports the tokens via `@import "@fanvue/ui/styles/theme.css"`, postcss resolves that against the package, and with the directory absent Turbopack falls back to a path outside the project root and refuses to compile.

Two changes, both scoped to `vite.config.ts`:

- a `copyStyles()` plugin that copies `theme.css` and `base.css` on `writeBundle`, so watch rebuilds are self-sufficient
- `emptyOutDir: false`, because `--watch` re-emits only what its incremental graph rebuilds — clearing the directory drops every artifact it does not rewrite, including entry points added since the watcher started

`pnpm build` still produces a complete `dist` either way.

## Verification

- `pnpm test` — 178 files / 4455 tests passing
- `pnpm build` after `rm -rf dist` emits `dist/styles/{theme,base}.css`, proving the plugin does the work rather than the leftover `cp`
- `pnpm size-limit` — 193.14 kB against the 220 kB budget

## Notes

The PR template's component checklist (story, a11y test, forwardRef, `src/index.ts` export) doesn't apply — this is build configuration and adds no public surface.

Typed `chore` deliberately, so release-please cuts no version for it. Local-dev only; nothing in the published package changes.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -9,12 +9,10 @@ const STYLE_FILES = ["theme.css", "base.css"];
 /**
  * Copies the design-token stylesheets into `dist/styles/`.
  *
- * These are consumed by eden via `@import "@fanvue/ui/styles/theme.css"`, which
- * postcss resolves against the package. The `build` npm script used to do this
- * with a trailing `cp`, so `vite build --watch` never produced them and every
- * rebuild left `dist/styles` missing — which breaks `pnpm start:localui`
- * (Turbopack then falls back to a path outside the project root and refuses to
- * compile). Doing it in the build itself keeps watch mode self-sufficient.
+ * eden imports these via `@import "@fanvue/ui/styles/theme.css"`, which postcss
+ * resolves against the package. Emitting them from the build rather than a
+ * trailing `cp` in the npm script keeps `vite build --watch` self-sufficient;
+ * without them `start:localui` fails to compile.
  */
 const copyStyles = (): Plugin => ({
   name: "fanvue-ui-copy-styles",
@@ -81,17 +79,9 @@ export default defineConfig({
       ],
     },
     cssCodeSplit: false,
-    // A watch rebuild re-bundles everything, but with `emptyOutDir` on it does so
-    // in two visible steps: vite clears `dist` at `renderStart`, before anything
-    // is written, then refills it as outputs finish. Polling through one rebuild
-    // shows `dist` go 1179 files, then 0, then 588, then back to 1179 — roughly
-    // three seconds where a reader sees an empty or half-written `dist`. That is
-    // the window that breaks `pnpm start:localui`, since eden resolves against
-    // `dist` and 500s when an import is missing. Leaving the directory in place
-    // means a rebuild only overwrites, so there is never a gap. Builds then never
-    // delete, which would let a renamed or removed component linger in a local
-    // `dist` and get published (`files: ["dist"]` + `prepublishOnly`), so the
-    // `build` script does the cleaning explicitly with `rm -rf dist`.
+    // Off so a watch rebuild never leaves `dist` empty mid-flight: vite clears it
+    // at `renderStart` and only refills at the end, and `start:localui` 500s in
+    // that window. Builds then never delete, so `build` does `rm -rf dist`.
     emptyOutDir: false,
     sourcemap: true,
     minify: false,
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -18,7 +18,9 @@ const STYLE_FILES = ["theme.css", "base.css"];
  */
 const copyStyles = (): Plugin => ({
   name: "fanvue-ui-copy-styles",
-  writeBundle() {
+  // `closeBundle`, not `writeBundle`: the latter fires once per output and we
+  // declare two (es into `dist`, cjs into `dist/cjs`), so it would copy twice.
+  closeBundle() {
     const outDir = resolve(import.meta.dirname, "dist/styles");
     mkdirSync(outDir, { recursive: true });
     for (const file of STYLE_FILES) {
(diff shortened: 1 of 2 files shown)


vite.config.ts
```diff
@@ -54,6 +79,12 @@ export default defineConfig({
       ],
     },
     cssCodeSplit: false,
+    // `vite build --watch` re-emits only what its incremental graph rebuilds, so
+    // clearing the directory drops every artifact it does not rewrite — entry
+    // points added since the watcher started, and dist/styles. eden then fails to
+    // resolve them and the page 500s. Leaving the directory in place keeps a
+    // watch rebuild additive; `pnpm build` still produces a complete dist.
+    emptyOutDir: false,
```
alfonsovo-fv, line 85: Bit of context on this flag, because it has a wider blast radius than it looks.

`emptyOutDir` is what makes vite wipe `dist` before it writes. Turning it off means builds only ever add and overwrite, they never remove. I tested it: I dropped a junk file at `dist/components/ZZZStale/ZZZStale.mjs`, ran a full `vite build`, and it was still sitting there afterwards.

So once we delete or rename a component, its old file lives on in your local `dist` forever. Three things make that bite here: `preserveModules` means one output file per source file, `files: ["dist"]` in package.json means whatever is in `dist` is what gets published, and `prepublishOnly` runs `pnpm build`. CI is safe because it always starts from a fresh checkout, but a local publish, or a long lived local `dist` that eden reads from, is not.

What I would do is keep the flag false, since it genuinely helps watch mode, and clean explicitly in the script instead:

```
"build": "rm -rf dist && vite build"
```

That gets you both: watch stays additive, and a real build always starts from empty. It also lets the trailing `cp` go, which covers my other point.

vite.config.ts
```diff
@@ -54,6 +79,12 @@ export default defineConfig({
       ],
     },
     cssCodeSplit: false,
+    // `vite build --watch` re-emits only what its incremental graph rebuilds, so
+    // clearing the directory drops every artifact it does not rewrite — entry
+    // points added since the watcher started, and dist/styles. eden then fails to
+    // resolve them and the page 500s. Leaving the directory in place keeps a
+    // watch rebuild additive; `pnpm build` still produces a complete dist.
```
alfonsovo-fv: The reasoning here is not quite right, and since this is a comment someone will read in six months and trust, I think it is worth correcting.

Rollup's watch mode does not do partial output. Every rebuild re-bundles the whole thing, it just reuses cached module transforms to go faster. I checked on main by diffing the list of files in `dist` before and after a rebuild and it was identical, 1179 files both times. Same for the entry points line, `lib.entry` is static config so adding one needs a watcher restart regardless.

The flag does help, just for a different reason: timing. With `emptyOutDir` on, vite clears `dist` at `renderStart`, which happens before anything is written, then refills it at the end. I polled through one rebuild and `dist` went 1179 files, then 0, then 588, then back to 1179. So there are roughly 3 seconds of every rebuild where anything reading `dist` sees nothing, or half a build. That window is what breaks eden, and that is the thing worth writing down here.

Worth internalising the general shape of this one: the change was right, the story about why was wrong, and only the story ends up in the codebase.

vite.config.ts
```diff
@@ -1,8 +1,32 @@
...
+ * These are consumed by eden via `@import "@fanvue/ui/styles/theme.css"`, which
+ * postcss resolves against the package. The `build` npm script used to do this
+ * with a trailing `cp`, so `vite build --watch` never produced them and every
+ * rebuild left `dist/styles` missing — which breaks `pnpm start:localui`
+ * (Turbopack then falls back to a path outside the project root and refuses to
+ * compile). Doing it in the build itself keeps watch mode self-sufficient.
+ */
+const copyStyles = (): Plugin => ({
+  name: "fanvue-ui-copy-styles",
+  writeBundle() {
```
alfonsovo-fv: Small one, not blocking.

`writeBundle` fires once per output, and we declare two of them in `rollupOptions.output`, the es build into `dist` and the cjs build into `dist/cjs`. So this copy runs twice on every build. `closeBundle` runs once, after all outputs are finished, so it is the more natural hook when the work is not tied to a specific bundle.

No harm as it stands, copying two small files twice costs nothing, but the writeBundle vs closeBundle distinction is a handy one to have in your head.

alfonsovo-fv (commented):
Nice fix, and the diagnosis is right. I reproduced it before reviewing: on main, once `vite build --watch` finishes its first build, `dist/styles` is simply not there, because vite clears `dist` and the `cp` only lives in the npm script. With your plugin the files are there and stay there through every rebuild. I also ran a plain `vite build` with your config and no `cp`, and it emits both files, so the plugin is definitely doing the work.

One thing to fix though: the `cp` is still in `package.json` line 55. Worth removing, both so there is a single source of truth, and because while it is there the "pnpm build emits dist/styles" check in your description cannot actually tell us whether the plugin or the `cp` produced them. It happens to be the plugin, I checked separately, but that test would pass either way. General habit worth picking up: when you replace a step, delete the old one in the same PR, otherwise the thing you are verifying is not the thing you changed.

Left three inline comments. The `emptyOutDir` one is the only one I would want sorted before this merges, the other two are context.

vite.config.ts
jacksonford-07: Fixed in `3c99426`. Went with your suggestion exactly: `"build": "rm -rf dist && vite build"`, so watch stays additive and a real build always starts from empty.

Ran your test rather than assuming: planted `dist/components/ZZZStale/ZZZStale.mjs`, ran `pnpm build`, and it is gone. Also checked `prepublishOnly` really is `pnpm run build` (package.json:72), since that is what makes this cover the publish path and not just the local one.

That also let the trailing `cp` go, which closes your other point.

vite.config.ts
jacksonford-07: You were right, and the distinction mattered more than the flag did. Rewritten in `3c99426` around the timing window instead of partial output.

Trimmed again in `df774fc` down to three lines. The measurements you gave, 1179 then 0 then 588, live in this thread and the PR rather than the config, on the principle that the file should carry what the next reader needs and not the investigation that produced it. What stayed is the bit that stops someone undoing it: that vite clears `dist` at `renderStart` and only refills at the end, and that `build` therefore does the cleaning with `rm -rf dist`.

The reason it was worth correcting: the old comment implied watch mode emits partially, so the obvious fix for the next reader would be to turn `emptyOutDir` back on and re-add the `cp`, which puts the bug straight back.

vite.config.ts
jacksonford-07: Switched to `closeBundle` in `3c99426`.

Verified the difference rather than taking it on trust: logged a line in the hook and it fires twice on `writeBundle`, once on `closeBundle`. Your read of the two outputs was exactly right.

alfonsovo-fv (approved):
All three addressed, approving. Verified the build end to end rather than reading it: planted a junk file in `dist`, ran `rm -rf dist && vite build`, and got `dist/styles/{base,theme}.css` present with the stale file gone. So the styles come from the plugin, and a real build starts from empty again.

`closeBundle` over `writeBundle` with the reason in a comment is right, and the `emptyOutDir` comment now says the thing that is actually true, the mid-flight empty window rather than partial re-emission. That was the one that mattered, since the wrong version would have been believed by whoever read it next.

The trimmed comments are better too. Same information, less of it.

PR #621, opened 2026-07-30
refactor(utils): move usePrefersReducedMotion out of CyclingText

jacksonford-07:
Moves `usePrefersReducedMotion` from `src/components/CyclingText/` to `src/utils/`, so it can be shared.

## Why

The hook is a generic `prefers-reduced-motion` media-query listener with nothing `CyclingText`-specific in it, but it lived inside that component's folder. `AnimatedNumber` needs the same behaviour, and importing across component folders is the wrong direction.

Pure move: the file's contents are unchanged (git reports it as a 100% rename), and the only other edit is `CyclingText`'s import path.

## Verification

- `CyclingText` suite — 2 files / 27 tests passing
- No references remain to the old path

## Notes

Typed `refactor` deliberately, so release-please cuts no version. The hook was never exported from `src/index.ts`, so there is no public API change — nothing in the published package moves.


src/components/CyclingText/CyclingText.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from "react";
 import { cn } from "../../utils/cn";
+import { usePrefersReducedMotion } from "../../utils/usePrefersReducedMotion";
```
alfonsovo-fv, line 3: Commenting here because the moved file itself has no diff lines to attach to.

A suggestion for a future PR, definitely not this one, since this should stay a pure move. The hook subscribes to something outside React (a media query) using `useState` plus `useEffect`. React 18 added `useSyncExternalStore` for exactly that shape, and it is worth knowing about:

```ts
const subscribe = (onChange: () => void) => {
  const mq = window.matchMedia("(prefers-reduced-motion: reduce)");
  mq.addEventListener("change", onChange);
  return () => mq.removeEventListener("change", onChange);
};

export function usePrefersReducedMotion(): boolean {
  return React.useSyncExternalStore(
    subscribe,
    () => window.matchMedia("(prefers-reduced-motion: reduce)").matches,
    () => false, // server snapshot
  );
}
```

Why it is nicer: the current version always renders `false` first and only corrects itself after the effect runs, so someone with reduced motion on can catch one animated frame before it settles. `useSyncExternalStore` reads the real value during render, and it is also the version that behaves correctly under concurrent rendering, where an effect based subscription can read a stale value.

The current code is not wrong and the separate server snapshot argument keeps SSR safe either way. Just a pattern worth having in your head next time you reach for useState plus useEffect to track something the browser owns.

alfonsovo-fv (approved):
Approving, this is a clean move. I checked the things worth checking: it really is a content identical rename, `CyclingText.tsx` was the only importer anywhere in the repo so nothing is left pointing at the old path, and the hook was never exported from `src/index.ts` so `refactor` is the right type.

Good instinct splitting this out rather than bundling it into the AnimatedNumber work. A pure move is trivial to review, a move plus new behaviour is not, and if something breaks later a one line rename commit is much easier to bisect past. Worth making a habit of.

One thing to consider, not blocking. Now that this is shared infrastructure rather than one component's private helper, it has no test of its own, and nothing in the repo covers the case where the user actually has reduced motion switched on. The CyclingText test is even named "completes a transition step without relying on matchMedia". Two small tests would lock both branches down: one stubbing `window.matchMedia` to return `matches: true`, and one firing a `change` event to check the listener updates state. `src/utils` is honestly mixed on coverage already, so you are not out of pattern, but a hook two components depend on is a fair place to start being strict.

src/components/CyclingText/CyclingText.tsx
jacksonford-07: Agreed, keeping this one a pure move.

Good pointer on `useSyncExternalStore`, and you are right about the flash: the current shape returns `false` until the effect runs, so someone with reduced motion on can catch a frame before it settles.

Taking it in a follow-up PR rather than here, using the version you sketched. I will link it back to this thread.

jacksonford-07: Follow-up is #630.

One deviation from your snippet worth flagging: I kept the `matchMedia` guard. `getSnapshot` runs during render and jsdom does not implement `matchMedia`, `src/test/setup.ts` polyfills `ResizeObserver` but not that one, so the unguarded version throws rather than degrades. That is also why the existing test is called "completes a transition step without relying on matchMedia".

Also checked the new test actually distinguishes the two implementations: it fails on the effect-based version and passes on this one, and with a faithful stub, `matches` as a live getter rather than a snapshot, it is the only one of the five that changes verdict. Your read was exactly right that the first render is the defect; change handling was already fine.

PR #622, opened 2026-07-30
feat(Chip): add an outlined variant

jacksonford-07:
Adds an `outlined` prop to `Chip`: the unselected state draws a 1px `buttons-chip-default` stroke over a transparent background instead of filling with that same token.

## Why

The Figma `V2 Insights Chips` treatment. It's used where a row of chips sits directly over a chart — a row of solid fills competes with the plotted series for attention, while outlines recede and let the selected chip read as the odd one out.

Additive and defaulted `false`, so every existing chip renders byte-identically. The selected state is untouched (still filled), as are the `dark` and `dotted` variants.

## One detail worth reviewing

The border is declared in **both** the selected and unselected states, changing only its colour — going transparent once selected — rather than being dropped on selection.

It can't be dropped: `border-box` constrains only explicitly sized axes, and a chip's width is intrinsic. Removing the border on selection would narrow the chip by 2px and reflow the whole row on every click. Height is safe either way, because `h-8` is explicit.

## Verification

- `Chip` suite — 83 tests passing, including new coverage for the variant and for the no-reflow behaviour above
- Two stories added, `Outlined` and `OutlinedSelected`, plus the prop wired into the Storybook controls

## Hover, and where its values come from

Raised in review: the hover rule never excluded `outlined`, so it inherited the filled chip's `hover:bg-buttons-chip-hover`. Against the real Figma values that is a **20% fill sitting over a 10% stroke**, so the outline disappeared into the fill at the moment of interaction, which works against the reason the variant exists.

Both hover values are existing **atomics** — no new tokens, and no raw primitives:

| | token | resolves to |
| --- | --- | --- |
| resting stroke | `buttons-chip-default` | `#1515151a` (10%) |
| hover stroke | `buttons-chip-hover` | `#15151533` (20%) |
| hover surface | `interaction-hover` | `#1515150d` (5%) |

`buttons-chip-hover` is design's one-step darkening for chips. As a *fill* it swallows a 10% stroke; moved onto the *border* it is the same step applied where the variant actually reads. `interaction-hover` is the atomic hover surface, and at 5% it cannot compete with a 20% stroke — the same relationship `dotted` already relies on.

**On design provenance, stated plainly:** the resting state is confirmed against Figma. The chips live in the Insights designs under `V2 Insight Card` (stable ancestor `8279:35573`; the chip instances themselves have non-persistent component-internal IDs), and they use `Buttons/Chip/Default` for the outline and `Buttons/Chip/Active` when selected, which is what this PR does. **Hover is not drawn anywhere reachable** — static product frames carry no hover state, and `V2 Insights Chips` in the library has no hover variant. So the hover values are derived from the atomics rather than read from a spec. Flagging that rather than implying parity.

Follow-up for design (Casp): draw hover, pressed and disabled on `V2 Insights Chips` and add the hover-stroke token to the atomics, so this stops being inferred.

## Verification

- Rendered the awkward combinations and dumped the real class strings rather than reading the conditions. `outlined selected disabled` did come out `bg-buttons-chip-active` + `text-content-disabled`; it now takes the muted fill and keeps `border-transparent` to hold width.
- Confirmed Tailwind actually emits `hover:border-buttons-chip-hover` — this token had never been used as a border anywhere, and an ungenerated utility fails silently.
- Real mouse hover in a browser: border 10% → 20%, background 0 → 5%, width unchanged at 52.8px so hover cannot reflow the row.
- Added `OutlinedInteractive`. Without it the hover was unreachable in Storybook, since both existing outlined stories pass no `onClick` — which is why this was never caught visually.
- 42 `Chip` tests, `typecheck` and `biome` clean.

## Split out

The `dotted` variant had the same 2px reflow bug (measured: 87.30px → 89.30px on toggle). Fixed separately in #631 rather than widening this PR.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -183,14 +183,14 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
           // Outlined darkens its stroke rather than filling. `buttons-chip-hover` is
           // the one-step darkening for chips, but as a fill it is 20% over a 10%
           // stroke, so it swallows the outline the variant exists to show. Moved onto
-          // the border, with the same 5% fill `dotted` uses under its darkened stroke.
+          // the border, over `interaction-hover`, the atomic hover surface.
           isInteractive &&
             !disabled &&
             !isDark &&
             !selected &&
             !dotted &&
             outlined &&
-            "hover:border-buttons-chip-hover hover:bg-neutral-alphas-50",
+            "hover:border-buttons-chip-hover hover:bg-interaction-hover",
           isInteractive &&
             !disabled &&
             !isDark &&
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -26,6 +26,15 @@ export interface ChipProps extends React.HTMLAttributes<HTMLElement> {
    * @default false
    */
   dotted?: boolean;
+  /**
+   * Draws the unselected state as a 1px `buttons-chip-default` outline over a
+   * transparent background instead of filling with that same token. This is the
+   * Figma `V2 Insights Chips` treatment, used where chips sit over a chart and a
+   * row of solid fills would compete with the plotted series. The selected state
+   * is unaffected — it stays filled — as are the `dark` and `dotted` variants.
+   * @default false
+   */
+  outlined?: boolean;
   /** Icon element displayed before the label. */
   leftIcon?: React.ReactNode;
   /** Icon element displayed after the label. */
@@ -64,6 +73,7 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
       disabled = false,
       leftDot = false,
       dotted = false,
+      outlined = false,
       leftIcon,
       rightIcon,
       notificationLabel,
@@ -131,7 +141,24 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
             selected &&
             dotted &&
             "border border-border-primary border-solid bg-inputs-inputs-primary text-content-primary",
-          !isDark && !selected && !dotted && "bg-buttons-chip-default text-content-primary",
+          !isDark &&
+            !selected &&
+            !dotted &&
+            !outlined &&
+            "bg-buttons-chip-default text-content-primary",
+          // Outlined: the same token as a 1px stroke rather than a fill. The border
+          // is declared in BOTH states and only changes colour, going transparent
+          // once selected. It cannot be dropped on selection: `border-box` constrains
+          // only explicitly sized axes, and a chip's width is intrinsic, so losing the
+          // border would narrow the chip by 2px and reflow the whole row on every
+          // click. Height is safe either way because `h-8` is explicit.
+          !isDark && !dotted && outlined && "border border-solid",
+          !isDark &&
+            !selected &&
+            !dotted &&
+            outlined &&
+            "border-buttons-chip-default bg-transparent text-content-primary",
+          !isDark && selected && !dotted && outlined && "border-transparent",
           // Dotted (non-selected): the dashed border is drawn via SVG (`dottedBorder`).
           // `group` lets that SVG react to hover/active. `asChild` keeps a CSS dashed
           // border fallback since the SVG is only rendered in the default path.
@@ -165,7 +192,9 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
           disabled && !isDark && "pointer-events-none text-content-disabled",
           // Solid (non-dotted) disabled chips get a muted fill; dotted ones stay
           // transparent with their dashed border (drawn via SVG).
-          disabled && !isDark && !dotted && "bg-buttons-chip-disabled",
+          disabled && !isDark && !dotted && !outlined && "bg-buttons-chip-disabled",
+          // Outlined stays unfilled when disabled; only its stroke greys out.
+          disabled && !isDark && !dotted && outlined && "border-buttons-chip-disabled",
           className,
         )}
         {...(isInteractive && {
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -178,7 +178,19 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
             !isDark &&
             !selected &&
             !dotted &&
+            !outlined &&
             "hover:bg-buttons-chip-hover",
+          // Outlined darkens its stroke rather than filling. `buttons-chip-hover` is
+          // the one-step darkening for chips, but as a fill it is 20% over a 10%
+          // stroke, so it swallows the outline the variant exists to show. Moved onto
+          // the border, with the same 5% fill `dotted` uses under its darkened stroke.
+          isInteractive &&
+            !disabled &&
+            !isDark &&
+            !selected &&
+            !dotted &&
+            outlined &&
+            "hover:border-buttons-chip-hover hover:bg-neutral-alphas-50",
           isInteractive &&
             !disabled &&
             !isDark &&
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -91,13 +91,19 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
     const isInteractive = !!onClick && !asChild;
     const Comp = asChild ? Slot : isInteractive ? "button" : "span";
     const isDark = variant === "dark";
+    // Precedence between the three treatments, named once rather than re-derived per
+    // rule: `dark` owns its own palette, `dotted` wins over `outlined`, and what is
+    // left is the normal chip, either filled or outlined.
+    const isDotted = !isDark && dotted;
+    const isSolid = !isDark && !dotted && !outlined;
+    const isOutlined = !isDark && !dotted && outlined;
 
     // The dashed border is drawn as an SVG so the dash length matches the design
     // spec (8/8 for the 40px square, 6/6 otherwise). CSS `border-dashed` only
     // renders browser-default dash lengths, which are too short. The stroke colour
     // is driven by `currentColor`, and `group-hover`/`group-active` react to the
     // chip's interactive states. Rendered only in the default (non-`asChild`) path.
-    const showDottedBorder = !isDark && dotted && !selected && !asChild;
+    const showDottedBorder = isDotted && !selected && !asChild;
     const dottedBorder = showDottedBorder ? (
       <svg
         aria-hidden="true"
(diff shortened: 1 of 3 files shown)


jacksonford-07:
**Storybook:** Components → Chip → **Outlined** and **OutlinedSelected**. The `outlined` control is also wired into the props panel, so you can toggle it against `selected` on any Chip story.

Chromatic is labelled on this PR, so its comment carries a hosted Storybook build plus before/after diffs for every existing Chip story — worth checking those, since the change edits the *unselected* fill condition that all 53 unrelated consumers already run.

src/components/Chip/Chip.tsx
```diff
@@ -165,7 +192,9 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
           disabled && !isDark && "pointer-events-none text-content-disabled",
           // Solid (non-dotted) disabled chips get a muted fill; dotted ones stay
           // transparent with their dashed border (drawn via SVG).
-          disabled && !isDark && !dotted && "bg-buttons-chip-disabled",
+          disabled && !isDark && !dotted && !outlined && "bg-buttons-chip-disabled",
+          // Outlined stays unfilled when disabled; only its stroke greys out.
+          disabled && !isDark && !dotted && outlined && "border-buttons-chip-disabled",
```
alfonsovo-fv: I think this one has a hole in it. Scoping line 195 to `!outlined` means the muted disabled fill never lands for an outlined chip, and nothing else overrides `bg-buttons-chip-active` from line 138. I rendered it to check, and `outlined selected disabled` comes out as:

```
bg-buttons-chip-active  text-content-disabled  border-buttons-chip-disabled
```

versus the plain equivalent:

```
bg-buttons-chip-disabled  text-content-disabled
```

So a disabled selected outlined chip is a solid black chip with grey disabled text sat on top of it. Not what you want, and the contrast will be poor.

Your comment says "outlined stays unfilled when disabled", which is true for the unselected case, but a selected chip is already filled, so it needs excluding:

```ts
disabled && !isDark && !dotted && (!outlined || selected) && "bg-buttons-chip-disabled",
disabled && !isDark && !dotted && outlined && !selected && "border-buttons-chip-disabled",
```

Selected then falls back to the normal disabled fill, and `border-transparent` from line 161 stays in place so the width still does not move. Worth adding a test for the selected plus disabled combination too, the current five only cover unselected there.

src/components/Chip/Chip.tsx
```diff
@@ -131,7 +141,24 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
...
+          // once selected. It cannot be dropped on selection: `border-box` constrains
+          // only explicitly sized axes, and a chip's width is intrinsic, so losing the
+          // border would narrow the chip by 2px and reflow the whole row on every
+          // click. Height is safe either way because `h-8` is explicit.
+          !isDark && !dotted && outlined && "border border-solid",
+          !isDark &&
+            !selected &&
+            !dotted &&
+            outlined &&
+            "border-buttons-chip-default bg-transparent text-content-primary",
```
alfonsovo-fv, line 161: Does the outlined chip want a fill on hover?

The hover rule further down does not exclude `outlined`, so an interactive outlined chip ends up with both `bg-transparent` and `hover:bg-buttons-chip-hover`. Those do not conflict, one is the base and one is the hover state, so tailwind-merge keeps both. Checking the tokens, `buttons-chip-default` is `neutral-alphas-100` and `buttons-chip-hover` is `neutral-alphas-200`, so on hover it fills with something darker than its own stroke and the outline effectively disappears into the fill.

Given the whole point is to not compete with the chart underneath, I would expect hover to darken the stroke rather than fill the shape. The `dotted` variant already does exactly that, via `buttons-chip-dotted-hover-stroke`. Worth checking against Figma, and either way it wants a hover condition of its own rather than inheriting the filled one by accident.

General thing worth taking from this one: when you add a variant by adding conditions, the risk is in the conditions you did not touch. Scanning the class list for every rule that does not mention your new flag is a quick way to find what you have silently inherited.

src/components/Chip/Chip.tsx
```diff
@@ -131,7 +141,24 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
...
+            !dotted &&
+            !outlined &&
+            "bg-buttons-chip-default text-content-primary",
+          // Outlined: the same token as a 1px stroke rather than a fill. The border
+          // is declared in BOTH states and only changes colour, going transparent
+          // once selected. It cannot be dropped on selection: `border-box` constrains
+          // only explicitly sized axes, and a chip's width is intrinsic, so losing the
+          // border would narrow the chip by 2px and reflow the whole row on every
+          // click. Height is safe either way because `h-8` is explicit.
+          !isDark && !dotted && outlined && "border border-solid",
```
alfonsovo-fv: Not blocking, a readability thought.

These four new rules all repeat `!isDark && !dotted`, and the component already carries a biome ignore for cognitive complexity, so it is worth not adding to it. Two derived booleans up near `isDark` would flatten it:

```ts
const isSolid = !isDark && !dotted && !outlined;
const isOutlined = !isDark && !dotted && outlined;
```

The rules then read `isOutlined && "border border-solid"`, `isOutlined && !selected && "..."` and so on. Same output, but the precedence between dark, dotted and outlined gets stated once in something with a name, instead of being re-derived in six places where one of them can quietly disagree with the others.

src/components/Chip/Chip.tsx
```diff
@@ -131,7 +141,24 @@ export const Chip = React.forwardRef<HTMLButtonElement, ChipProps>(
             selected &&
             dotted &&
             "border border-border-primary border-solid bg-inputs-inputs-primary text-content-primary",
-          !isDark && !selected && !dotted && "bg-buttons-chip-default text-content-primary",
+          !isDark &&
+            !selected &&
+            !dotted &&
+            !outlined &&
+            "bg-buttons-chip-default text-content-primary",
+          // Outlined: the same token as a 1px stroke rather than a fill. The border
```
alfonsovo-fv, line 152: Good catch, and thanks for writing the reasoning down rather than just fixing it. This is the kind of thing that gets "simplified" away by the next person otherwise.

Since you clearly have the eye for it: the `dotted` variant has the same bug today. Unselected dotted has no border, because the dash is drawn as an SVG, but line 143 adds `border border-border-primary border-solid` when selected. So toggling a dotted chip shifts the row by 2px, exactly what you are preventing here. Not for this PR, but worth raising a ticket while it is fresh.

One small thing on the matching test: "toggling cannot resize the chip" is asserting class names rather than measuring anything, and jsdom cannot do layout so that is genuinely the best available. I would name it after what it checks though, something like "keeps a border in both states", and leave the actual no-shift proof to Chromatic. Test names get read as promises, so it is worth keeping them to what the assertions can back up.

alfonsovo-fv (changes requested):
The shape of this is good. It is properly additive, the existing rules only gained `!outlined` guards so current consumers are untouched, and the story and test coverage are there. Your `border-box` reasoning in the description is also correct, which is a genuinely non-obvious detail: a border is only absorbed on an axis with an explicit size, so `h-8` swallows it vertically while the intrinsic width does not. Declaring the border in both states and only recolouring it is the right fix.

Rather than read the conditions I rendered the awkward combinations and dumped the real class strings, and two things fell out.

Requesting changes for one of them: `outlined selected disabled` comes out as a solid black chip with grey disabled text, because the muted disabled fill is now skipped for outlined chips and nothing overrides the active fill. Details inline on line 197, along with a suggested condition.

The second is a design question rather than a defect: an outlined chip still fills on hover, with a token darker than its own stroke, so the outline all but vanishes at the moment you interact with it. Inline on line 160. Worth checking Figma before merge, since it works against the reason the variant exists.

Two non-blocking notes as well, one readability and one about a pre-existing bug of the same kind you fixed here.

src/components/Chip/Chip.tsx
jacksonford-07: Fixed in `39d0df5`, and I went and read the design for this one. Summary: it confirms the resting state and does not specify hover at all.

Found the chips in the Insights mobile designs. Note for anyone else digging: the instances are nested inside `V2 Insight Card`, so their node IDs are component-internal and non-persistent, and the API rejects them. The stable ancestor is `8279:35573`.

What the design settles:

- The outline uses `Color/Buttons/Chip/Default`, the same token as the filled chip, which is what this PR already does. In the dark-mode frame that resolves to `#ffffff33`.
- Selected is a full fill of `Buttons/Chip/Active`, `#ffffff` in dark. Matches.
- Chips are 32px with an 8px gap (`Spacing/Group/chip-chip`).

What it does not settle: hover. Static product frames carry no hover state, and there is no Insights-specific hover token anywhere in the library. `search_design_system` does turn up a `Color/Buttons/Posts/Insights/*` family with a `Hover-Stroke`, but that sits under Buttons/Posts and none of it is referenced by these frames, so I do not think it is ours.

So the values here are derived, not read, and I would rather say so than imply parity. From the real tokens: resting stroke is 10%, the inherited hover fill was 20%, so the fill was darker than the stroke it sat over, which is the bug you spotted. I moved that same one-step darkening onto the border and used the 5% fill that `dotted` already uses under its darkened stroke. Verified in a browser with a real mouse hover: border goes 10% to 20%, background 0 to 5%, width unchanged so hover cannot reflow.

Two asks. Happy for you to accept the derivation for now, and separately we should get Casp to draw hover, pressed and disabled on `V2 Insights Chips` in the library and add the token to the atomics, so the next person is not inferring it either. I will raise that.

One thing I noticed while in there, unrelated to this thread but relevant to the variant: the design gives each chip a dot in its chart series colour, purple, blue, yellow, green. `leftDot` renders `bg-current`, so it always matches the label and cannot be per-series. Consumers can use `leftIcon` with their own coloured node instead, but if this row is the target use case then `leftDot` may want a colour prop. Worth its own ticket rather than widening this PR.

src/components/Chip/Chip.tsx
jacksonford-07: Good catch, fixed in `a14611f`. You were right that nothing overrode `bg-buttons-chip-active`.

Verified before changing anything by rendering the combination and dumping the class string, which came out as you said:

```
outlined selected disabled  ->  bg-buttons-chip-active  text-content-disabled  border-buttons-chip-disabled
```

Took your condition. It now renders `bg-buttons-chip-disabled` with `border-transparent` still in place, so it matches the plain disabled chip and the width does not move. Added the selected plus disabled test too, since as you noted the existing five only covered unselected.

src/components/Chip/Chip.tsx
jacksonford-07: Raised and fixed in #631 rather than just a ticket, since it was fresh.

Confirmed it first, because jsdom cannot do layout and you had reasoned it rather than measured it. In a browser the dotted chip goes 87.30px unselected to 89.30px selected, so exactly the 2px you predicted. Height stayed 32px throughout, since `h-8` is explicit and absorbs the border on a sized axis.

Measuring also caught something reasoning would not have: reserving the transparent border fixed the width, but the SVG was still sized to the padding box, which put the dash 1.5px in from the edge while the selected border sits at 0.5px. They were not on the same ring, so toggling would have visibly nudged the outline. The SVG now spans the border box and both land at 0.5px.

Also took your point on the test name. It is now `keeps a border in both states`, which is what the assertions can actually back, and the no-shift proof is left to Chromatic.

src/components/Chip/Chip.tsx
jacksonford-07: Fair, and I checked rather than eyeballing it: it is worse than when you wrote this. `!isDark` now appears 16 times and the `!dotted &&` chain 9 times, because both of my follow-up commits added more `!isDark && !selected && !dotted && outlined` conditions. So I have added to the thing you flagged, on a file that already carries the `noExcessiveCognitiveComplexity` ignore.

I have left it out of this PR to keep the diff to the review points that gate it, but I think you are right and it wants doing. Happy to fold it in here instead if you would rather it did not ship as-is, or take it as a follow-up with the 42 existing tests plus a class-string diff across every prop combination to prove the output is byte-identical. Your call, I do not have a strong preference on which PR it lands in.

jacksonford-07: Done in `5ab8411`. Went with `isDotted`, `isSolid` and `isOutlined`. Your two could not quite cover it on their own: the selected fill on line 138 applies to solid and outlined both, so that one reads `(isSolid || isOutlined) && selected`, and the disabled fill is now `disabled && (isSolid || (isOutlined && selected))`, which states that rule as the sentence it actually is.

`!isDark` goes from 16 occurrences to 4, and the `!dotted &&` chain from 9 to 2.

Proved the output did not move rather than trusting the tests: dumped the root and svg class strings for all 768 prop combinations, before and after, and diffed. Byte-identical.

One correction to the premise though, since I checked it. The complexity score goes **up**, 54 to 59, not down, because biome counts the three derived booleans too. So the ignore is still required and this does add to it slightly. Getting that number down would mean lifting the whole class list into a module-level function, which felt like a bigger change than this thread was asking for, so I left it.

The other half of your argument does hold, and it is the half that mattered: the precedence is stated once now, so the rules cannot disagree. That is exactly how `outlined selected disabled` broke, one hand-written prefix out of step with the rest. Happy to go further if you want the score down as well.

alfonsovo-fv (approved):
All four addressed, approving. I re-rendered the combinations rather than reading the diff:

```
outlined selected disabled  ->  bg-buttons-chip-disabled  border-transparent  (no bg-buttons-chip-active)
outlined interactive        ->  hover:border-buttons-chip-hover  hover:bg-interaction-hover
plain interactive           ->  hover:bg-buttons-chip-hover      (unchanged)
dotted interactive          ->  unchanged
```

So the disabled fill lands and keeps its transparent border for width, the hover darkens the stroke 10% to 20% over a 5% fill, and neither solid nor dotted moved. `interaction-hover` resolves to `neutral-alphas-50`, the same 5% `dotted` already uses, so that is consistent rather than a new number.

Two things worth saying back.

You are right about the complexity score and I was wrong to imply the refactor would help the ignore. Biome counting the derived booleans pushing it 54 to 59 is not something I checked before suggesting it. The half that mattered is the half you kept: the precedence is stated once, so the rules cannot drift out of step, which is exactly how `outlined selected disabled` broke. Not worth chasing the number.

And going to the Figma frame, then saying plainly that hover is derived rather than specified, is the right way to handle a gap in the design. Much better than picking a value and letting it read as spec. Please do raise the `V2 Insights Chips` states with Casp, and the `leftDot` colour point is a good catch, worth its own ticket.

#631 too: measuring 87.30px to 89.30px confirmed what I had only reasoned, and the SVG ring being on a different radius than the border is something I missed entirely. Good.

PR #623, opened 2026-07-30
feat(IconButton): add the field variant

jacksonford-07:
Adds a `field` variant to `IconButton`, carrying the input surface rather than a button one.

## Why

Figma draws an icon-only action inside a row of fields as a `V2 Input Fields` instance with just an icon in it — not as a button. On the agency insights toolbar, the download control sits immediately beside a Select and a date-range field, and a button-surfaced icon broke the row into two visual groups.

`field` takes the same border, background and hover as the Select trigger it lines up with, so the row reads as one control group.

## The one thing that isn't like the others

`field` is the **only variant that is never circular**. Every other variant is shaped by size (`rounded-xs` at 24, circular above); `field` takes the inputs' `rounded-sm` at every size, because it has to match the fields either side of it rather than the buttons elsewhere.

That asymmetry is deliberate and called out in the type's doc comment, so it isn't mistaken for an oversight later.

## Verification

- `IconButton` suite — 55 tests passing, including coverage for the variant's surface and its non-circular shape at every size
- `Field` story added and the variant registered in the Storybook `variant` control
- `e2e/iconButton.spec.ts` doesn't reference variants, so the Playwright suite is unaffected

Additive — every existing variant is untouched, so nothing renders differently.

### DIFF
--- a/src/components/IconButton/IconButton.test.tsx
+++ b/src/components/IconButton/IconButton.test.tsx
@@ -53,7 +53,8 @@ describe("IconButton", () => {
         );
         const button = screen.getByTestId("icon-button");
         expect(button).toHaveClass("rounded-sm");
-        expect(button).not.toHaveClass("rounded-full", "rounded-xs");
+        expect(button).not.toHaveClass("rounded-full");
+        expect(button).not.toHaveClass("rounded-xs");
         unmount();
       }
     });
--- a/src/components/IconButton/IconButton.test.tsx
+++ b/src/components/IconButton/IconButton.test.tsx
@@ -46,6 +46,27 @@ describe("IconButton", () => {
       expect(button).not.toHaveClass("rounded-xs");
     });
 
+    it("gives the field variant the inputs' radius at every size, never circular", () => {
+      for (const size of ["24", "40"] as const) {
+        const { unmount } = render(
+          <IconButton icon={<HomeIcon />} variant="field" size={size} aria-label="Download" />,
+        );
+        const button = screen.getByTestId("icon-button");
+        expect(button).toHaveClass("rounded-sm");
+        expect(button).not.toHaveClass("rounded-full", "rounded-xs");
+        unmount();
+      }
+    });
+
+    it("gives the field variant the input surface and border", () => {
+      render(<IconButton icon={<HomeIcon />} variant="field" size="40" aria-label="Download" />);
+      expect(screen.getByTestId("icon-button")).toHaveClass(
+        "bg-inputs-inputs-primary",
+        "border",
+        "border-border-primary",
+      );
+    });
+
     it("keeps bespoke variants circular at every size", () => {
       render(<IconButton icon={<HomeIcon />} variant="microphone" size="24" aria-label="Mic" />);
       expect(screen.getByTestId("icon-button")).toHaveClass("rounded-full");
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -230,7 +230,8 @@ export interface IconButtonProps extends React.ButtonHTMLAttributes<HTMLButtonEl
  * `contrast`, `messaging`, `navTray`, `tertiaryDestructive`, `stop`,
  * `microphone`) stay circular at all sizes. `field` is the exception that is
  * never circular — it takes the inputs' `rounded-sm` so it matches the fields it
- * sits beside.
+ * sits beside. Use it at `32`, `40` or `48`: those are the sizes `Select` offers,
+ * and the variant only earns its surface when it lines up with one.
  *
  * @example
  * ```tsx
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -27,7 +27,8 @@ export type IconButtonVariant =
   | "navTray"
   | "tertiaryDestructive"
   | "stop"
-  | "microphone";
+  | "microphone"
+  | "field";
 
 /** Icon button size in pixels. */
 export type IconButtonSize = "24" | "32" | "40" | "48" | "52" | "72";
@@ -184,6 +185,15 @@ const VARIANT_CLASSES: Record<IconButtonVariant, VariantClasses> = {
       "bg-buttons-primary-default text-content-primary-inverted hover:bg-buttons-brand-default hover:text-content-always-black not-disabled:active:bg-buttons-brand-default not-disabled:active:text-content-always-black",
     disabled: DISABLED_OPACITY,
   },
+  // Carries the input surface rather than a button one, so an icon-only action can
+  // sit flush in a row of fields (Figma draws these as a `V2 Input Fields` instance
+  // with only an icon inside). Border and hover match the Select trigger it lines up
+  // with, so the row reads as one control group.
+  field: {
+    default:
+      "border border-border-primary bg-inputs-inputs-primary text-content-primary not-disabled:hover:border-neutral-alphas-400",
+    disabled: DISABLED_TRANSPARENT,
+  },
 };
 
 function getVariantClasses(variant: IconButtonVariant, negative: boolean): string {
@@ -218,7 +228,9 @@ export interface IconButtonProps extends React.ButtonHTMLAttributes<HTMLButtonEl
  * Shape is size-driven for the standard variants: the `24` size is squared
  * (`rounded-xs`), every larger size is circular. Bespoke variants (`brand`,
  * `contrast`, `messaging`, `navTray`, `tertiaryDestructive`, `stop`,
- * `microphone`) stay circular at all sizes.
+ * `microphone`) stay circular at all sizes. `field` is the exception that is
+ * never circular — it takes the inputs' `rounded-sm` so it matches the fields it
+ * sits beside.
  *
  * @example
  * ```tsx
@@ -257,7 +269,13 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
           "relative inline-flex shrink-0 items-center justify-center focus-visible:outline-none",
           "cursor-pointer transition-all duration-150 ease-in-out disabled:cursor-default",
           "focus-visible:shadow-focus-ring",
-          SIZE_DRIVEN_SHAPE_VARIANTS.has(variant) && size === "24" ? "rounded-xs" : "rounded-full",
+          // `field` takes the inputs' radius at every size so it lines up with the
+          // fields around it; everything else keeps the existing size-driven shape.
+          variant === "field"
+            ? "rounded-sm"
+            : SIZE_DRIVEN_SHAPE_VARIANTS.has(variant) && size === "24"
+              ? "rounded-xs"
+              : "rounded-full",
           sizeVariants[size],
           getVariantClasses(variant, negative),
           className,
--- a/src/components/IconButton/IconButton.stories.tsx
+++ b/src/components/IconButton/IconButton.stories.tsx
@@ -153,6 +153,7 @@ const meta = {
         "tertiaryDestructive",
         "stop",
         "microphone",
+        "field",
       ],
     },
     size: {
@@ -695,5 +696,22 @@ export const AllIcons: Story = {
   ),
   parameters: {
     layout: "fullscreen",
+  },
+};
+
+/**
+ * Carries the input surface rather than a button one, so an icon-only action can sit
+ * flush in a row of fields — Figma draws these as a `V2 Input Fields` instance with
+ * only an icon inside. Its border and hover match the Select trigger it lines up
+ * with, so the row reads as one control group.
+ *
+ * The only variant that is never circular: it takes the inputs' `rounded-sm` at
+ * every size, where the other variants are shaped by size.
+ */
+export const Field: Story = {
+  args: {
+    variant: "field",
+    size: "40",
+    icon: <HomeIcon />,
   },
 };


jacksonford-07:
**Storybook:** Components → IconButton → **Field**, and `field` is in the `variant` control on every other IconButton story.

The thing to look at in Chromatic's diffs isn't the new variant, it's the **existing** ones: `field` adds a branch to the shared shape ternary that every variant flows through, across 194 unrelated consumers. If a radius shifted somewhere unexpected, that's where it shows.

src/components/IconButton/IconButton.tsx
```diff
@@ -184,6 +185,15 @@ const VARIANT_CLASSES: Record<IconButtonVariant, VariantClasses> = {
...
     disabled: DISABLED_OPACITY,
   },
+  // Carries the input surface rather than a button one, so an icon-only action can
+  // sit flush in a row of fields (Figma draws these as a `V2 Input Fields` instance
+  // with only an icon inside). Border and hover match the Select trigger it lines up
+  // with, so the row reads as one control group.
+  field: {
+    default:
+      "border border-border-primary bg-inputs-inputs-primary text-content-primary not-disabled:hover:border-neutral-alphas-400",
+    disabled: DISABLED_TRANSPARENT,
```
alfonsovo-fv, line 195: This is the one place the variant stops matching the thing it is matching.

`DISABLED_TRANSPARENT` is only `disabled:text-content-disabled`, so a disabled `field` button greys its icon but keeps the border and surface at full strength. The Select beside it does disabled with `opacity-50` (`Select.tsx` line 152), so it fades out entirely.

Which means in a row where both are disabled you get a crisp bordered icon button next to a faded Select, the same "two visual groups" problem the variant exists to fix. `DISABLED_OPACITY` is already in this file and would line them up:

```ts
disabled: DISABLED_OPACITY,
```

This is one of the two states I would want design to actually draw rather than either of us inferring, per my note on the PR.

Related, and the second one: `field` has no `not-disabled:active:` rule, where every other variant in this map does. The Select's equivalent is `data-[state=open]:border-neutral-alphas-400`. Is there meant to be no pressed feedback at all on a download button?

alfonsovo-fv (commented):
This one holds up well. I checked `field` against the Select trigger it is supposed to line up with rather than taking the description's word for it, and it matches exactly: `rounded-sm`, `border-border-primary`, `bg-inputs-inputs-primary`, and `hover:border-neutral-alphas-400` are the same four decisions the Select makes on lines 143, 148 and 151. The never-circular exception is genuinely justified and you documented it in three places, which is the right amount for something that will look like a mistake to the next reader. I also checked `negative`: `field` is not in `NEGATIVE_AWARE_VARIANTS`, so the prop is ignored rather than producing a broken half-state. Good.

Three comments inline, none of them blocking: the disabled treatment does not match the Select, there is no pressed state, and one of the new test assertions asserts nothing.

One process thing worth doing beyond the code. Right now this variant exists in our library but not in the design system, and the justification for it is "Figma draws this as a V2 Input Fields instance with an icon in it". That is a workaround being described as a spec. Please push design to add it properly as an IconButton variant in the Figma library, with its own hover, pressed and disabled states drawn. Two reasons it matters: you currently have to infer the disabled and pressed treatments, which is exactly where the two comments below come from, and until it is a real component in the library the next person solving the same problem will invent their own version of it rather than finding yours. Worth raising in the design channel and linking the ticket back here.

src/components/IconButton/IconButton.test.tsx
```diff
@@ -46,6 +46,27 @@ describe("IconButton", () => {
...
     });
 
+    it("gives the field variant the inputs' radius at every size, never circular", () => {
+      for (const size of ["24", "40"] as const) {
+        const { unmount } = render(
+          <IconButton icon={<HomeIcon />} variant="field" size={size} aria-label="Download" />,
+        );
+        const button = screen.getByTestId("icon-button");
+        expect(button).toHaveClass("rounded-sm");
+        expect(button).not.toHaveClass("rounded-full", "rounded-xs");
```
alfonsovo-fv: This line does not do what it looks like.

jest-dom's `toHaveClass` is a subset check, so negating it with two arguments only fails when **both** classes are present. I ran it to be sure: an element with `className="rounded-full"` passes `.not.toHaveClass("rounded-full", "rounded-xs")` quite happily.

Two separate assertions is what you want:

```ts
expect(button).not.toHaveClass("rounded-full");
expect(button).not.toHaveClass("rounded-xs");
```

The test is still sound as it stands, because tailwind-merge leaves only one `rounded-*` class and line 55 would catch a wrong one. But this line is currently free of charge, and the same pattern in a test without that backstop would pass while broken.

Rule of thumb worth keeping: one class per negated assertion. A positive `toHaveClass` with several arguments is fine, that is the subset check doing what you expect. It is only the negation that surprises you.

src/components/IconButton/IconButton.tsx
```diff
@@ -257,7 +269,13 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
           "relative inline-flex shrink-0 items-center justify-center focus-visible:outline-none",
           "cursor-pointer transition-all duration-150 ease-in-out disabled:cursor-default",
           "focus-visible:shadow-focus-ring",
-          SIZE_DRIVEN_SHAPE_VARIANTS.has(variant) && size === "24" ? "rounded-xs" : "rounded-full",
+          // `field` takes the inputs' radius at every size so it lines up with the
+          // fields around it; everything else keeps the existing size-driven shape.
+          variant === "field"
```
alfonsovo-fv, line 275: This reads fine, and I like that `field` is checked first so the existing size-driven path is left intact rather than reshuffled to accommodate it. Cheapest possible diff for the 194 consumers flowing through here.

A note on the API rather than this line. Select supports `32 | 40 | 48`, but `field` inherits all six IconButton sizes, and the test exercises `24`. So the type allows `field` at 52 or 72, sizes it can never actually sit beside a field at. Nothing to fix here, the component is not set up for per-variant size types and it would not be worth introducing that for one variant. But it is worth noticing when a variant's meaningful inputs are narrower than its type says, because tests written against the wider range end up documenting combinations the design never defined. If Figma only draws this at 40, saying so in the doc comment costs nothing.

src/components/IconButton/IconButton.test.tsx
jacksonford-07: Confirmed and fixed in 062f8fe. I ran your case before changing anything:

```
✓ OLD form: passes despite rounded-full being present
× NEW form: fails as it should
```

Same element, `className="rounded-full"`, only the assertion differs. So the two-argument negation was reporting nothing at all.

Split into two, and you were right that the test was still sound: line 55 asserts `rounded-sm` positively and tailwind-merge collapses to one `rounded-*`, so a wrong radius was always going to be caught. The assertion was decoration rather than a hole.

Took the rule of thumb wider and grepped for the pattern. Two more on `main`, both pre-existing rather than from this branch:

- `Chart.test.tsx:259`, three arguments
- `Chart.test.tsx:280`, two arguments

Left them alone since they are outside this PR, but 259 is the one I would look at, because three arguments means it only fails when all three are present and there is no positive assertion beside it doing the real work. Happy to pick those up separately if you want them.

src/components/IconButton/IconButton.tsx
jacksonford-07: Took the doc-comment half of this in 2d34716, since that part did not need design:

```
Use it at `32`, `40` or `48`: those are the sizes `Select` offers, and the
variant only earns its surface when it lines up with one.
```

Kept it to what the code already proves rather than what Figma draws, because `SelectSize` is `"48" | "40" | "32"` at `Select.tsx:9` and that is the real constraint. `52` and `72` remain typed but now read as a mistake rather than an option.

Left `24` in the test loop deliberately. It is not asserting that `24` is a supported size, it is asserting the radius rule holds at the one size where every other variant switches to `rounded-xs`. Drop it and the test stops covering the case most likely to break.

Your wider point stands though, and it is the one I will remember: the type says six sizes, the design defines three, and nothing in between reconciles them. Per-variant size types for one variant is not worth it, so the doc line is the cheap version of that.

src/components/IconButton/IconButton.tsx
jacksonford-07: Agreed on the diagnosis, holding on the fix, because both halves of this are the design question rather than a call either of us should make in the diff.

Confirmed what you described. `field` uses `DISABLED_TRANSPARENT`, which is only `disabled:text-content-disabled`, so the border and surface stay at full strength while the icon greys. `Select.tsx:152` fades the whole trigger with `opacity-50`. Disable a row and you get exactly the split you describe, which is the thing the variant exists to prevent.

`DISABLED_OPACITY` would line them up and it is already in the file, so it is a one-line change whenever we get an answer. I would rather not apply it on inference, because the two options are not equivalent: `opacity-50` fades the border too, which is what matches Select, but it also fades the icon twice if `text-content-disabled` ever comes back. Worth having drawn rather than guessed.

Same for the missing `not-disabled:active:` rule. You are right that `field` is the only entry in `VARIANT_CLASSES` without one. Select's equivalent is `data-[state=open]:border-neutral-alphas-400`, which is an open-state border rather than a press, so there is no obvious thing to copy. A download button almost certainly should not be the one control in the row with no pressed feedback, but what it should do is a design answer.

Taking both to Casp alongside the chart-yellow contrast question from #629. Will bring the answer back here rather than merging around it.

jacksonford-07: One more mismatch in the same state, found while checking the above: `Select.tsx:152` is `cursor-not-allowed opacity-50`, so the cursor differs too, not just the fade. `field` inherits the default cursor when disabled.

Same one-line fix and the same question, so folding it in rather than opening a thread. Three things to line up: fade, cursor, pressed state.

src/components/IconButton/IconButton.tsx
jacksonford-07: Went and tried the type-level version of this rather than leaving it as a note, because "constrain `field` to the sizes it can sit beside" sounds like it should be cheap. It is not, and the reason is worth writing down.

The discriminated union works:

```ts
export type IconButtonProps = IconButtonBaseProps &
  (
    | { variant: "field"; size?: "32" | "40" | "48" }
    | { variant?: Exclude<IconButtonVariant, "field">; size?: IconButtonSize }
  );
```

Two errors across the whole DS. One of them is my own test at `IconButton.test.tsx:52`, caught for using `24`, so the constraint does bite on real code. The other is `AudioRecordButton.tsx:54`, which forwards a `variant?: IconButtonVariant` straight through, and TS cannot correlate `variant` with `size` across a spread. Fixable, and only one site.

The part that kills it is downstream. `Omit` over a union collapses the discriminant, so the constraint evaporates the moment a consumer wraps the props:

```ts
const a: Props          = { icon: 1, variant: "field", size: "24" };  // error, good
const b: Omit<Props,"icon"> = { variant: "field", size: "24" };       // no error
```

I ran that. `a` fails, `b` passes.

`Omit<IconButtonProps, ...>` is the dominant pattern in eden. `MessageActionButton`, `SendMessageButton`, `StyledMenuButton`, `RemoveButton`, `MediaPreviewOverlayButton`, `NavExpandButton` are all built that way. So we would ship a type that constrains `field` in the DS and quietly stops constraining it at exactly the call sites most likely to get it wrong, while breaking pass-through wrappers that are perfectly correct today. Worse than the honest wide type, because it reads as a guarantee.

To the "regression elsewhere" worry, for the record: there is no path for this. `variant === "field"` is checked first at `IconButton.tsx:275` and every other variant falls through to the original size-driven expression unchanged. The looseness is a misuse risk on `field` itself, not a regression risk on the circular variants.

So the doc line stays the answer, and it is a better one than I thought when I wrote it. If we ever want this enforced properly it wants to be a separate exported component with its own narrow size type, not a union on the shared props.

jacksonford-07:
Threads all answered. Summary of where this landed:

**Fixed**
- `062f8fe` splits the negated radius assertion. Your case reproduced exactly: same element, `className="rounded-full"`, old form passes and new form fails. Two more instances of the pattern live on `main` at `Chart.test.tsx:259` and `:280`, left alone as out of scope.
- `2d34716` names `32 | 40 | 48` in the doc comment as the sizes `field` is for, matching `SelectSize`.

**Investigated and deliberately not changed**

I tried the discriminated union that would enforce those sizes. It works inside the DS, and it caught my own test for using `24`. But `Omit` over a union collapses the discriminant, and `Omit<IconButtonProps, ...>` is how eden builds six of its icon buttons, so the constraint disappears at the wrapping call sites while breaking correct pass-through wrappers. Detail in the thread. The wide type plus the doc line is the honest version.

**Waiting on design, not on you**

The disabled treatment and the missing pressed state, both as you called them. Confirmed `Select.tsx:152` is `cursor-not-allowed opacity-50` against `field`'s icon-only greying, so fade and cursor both differ, and `field` is the only entry in `VARIANT_CLASSES` with no `active:` rule. With Casp now, alongside the chart-yellow contrast from #629.

Both are one-liners in `VARIANT_CLASSES` once answered. Happy either way on whether that lands here or in a follow-up, but I did not want to guess two states in the diff.

jacksonford-07:
To firm up my "happy either way" above, since it leaves you guessing: my recommendation is merge this and take the two states as a follow-up. Do not hold it for Casp.

`variant="field"` has zero usages across `eden/src` and `fadmin/src`. Nothing consumes it yet, so a disabled state that does not match Select cannot regress anything today, and the design answer gates adoption rather than availability. Whoever builds the first field row has to get that state right before they ship and will be looking straight at it.

Against that, holding costs rebases. `main` has moved four times today and this branch is already behind.

Both fixes are one-liners in `VARIANT_CLASSES` once we have an answer, so the follow-up is about five lines. Yours if you would rather they land here instead, and I will wait.

alfonsovo-fv (approved):
Approving. The negated assertion is split and you verified the old form passed while the new one fails, and the size doc is grounded in `SelectSize` rather than in what Figma happens to draw, which is the better constraint to cite.

Holding the disabled fix for design is the right call, and you found a third mismatch in the same state (cursor) that I had missed. Nothing here should wait on that.

PR #624, opened 2026-07-30
feat(TabsList): add flushLeft and scrollable

jacksonford-07:
Adds two independent, opt-in props to `TabsList`. Both default `false`, so every existing tab bar is untouched.

## `flushLeft`

Drops the leading padding on the first tab so its label sits flush with the list's left edge, lining up with page content above or below the bar — a heading, for instance — rather than being indented by the tab's own padding. Spacing between tabs is unaffected.

## `scrollable`

Scrolls the row horizontally when the tabs don't fit, instead of letting them compress. By default a tab may shrink until its label ellipsises, which on a narrow viewport can truncate *every* label at once; here each tab keeps its natural width and the overflow scrolls. The scrollbar is hidden — the clipped tab at the edge is the affordance.

One consequence worth knowing: because the tabs become content-sized, they no longer spread to fill the row, so a set that *does* fit sits left-aligned as with `alignLeft`.

## Reviewing the diff

**Use `?w=1`.** GitHub reports ~248 changed lines in `TabsList.tsx`, but `git diff -w` gives **48 insertions / 2 deletions** — the rest is reindentation from the new destructuring wrapper. Ignoring whitespace makes this a small diff.

## Verification

- `Tabs` suite — 34 tests passing, with new coverage for both props including the no-compress behaviour
- Stories added for both

### DIFF
--- a/src/components/Tabs/TabsList.tsx
+++ b/src/components/Tabs/TabsList.tsx
@@ -115,9 +115,18 @@ export const TabsList = React.forwardRef<
         indicator.style.background =
           "linear-gradient(90deg, var(--color-buttons-tertiary-default) 0%, var(--color-tab-active) 50%, var(--color-buttons-tertiary-default) 100%)";
         const textSpan = activeTab.querySelector("span");
-        const textWidth = (textSpan ?? activeTab).getBoundingClientRect().width;
-        const tabCenter = activeTab.offsetLeft + activeTab.offsetWidth / 2;
-        const indicatorLeft = tabCenter - textWidth / 2;
+        const target = textSpan ?? activeTab;
+        const textWidth = target.getBoundingClientRect().width;
+        // Anchor to where the label actually sits, not to the tab's centre. The two
+        // only coincide when the tab's horizontal padding is symmetric, which
+        // `flushLeft` breaks by dropping `pl` on the first tab: the tab's centre then
+        // sits `pr / 2` to the right of the label's. Measured as a difference of two
+        // rects so it stays correct while a `scrollable` list is scrolled.
+        const textOffsetInTab =
+          target === activeTab
+            ? 0
+            : target.getBoundingClientRect().left - activeTab.getBoundingClientRect().left;
+        const indicatorLeft = activeTab.offsetLeft + textOffsetInTab;
         indicator.style.inset = "auto auto 0 0";
         indicator.style.height = "1px";
         indicator.style.width = `${textWidth}px`;
--- a/src/components/Tabs/TabsList.tsx
+++ b/src/components/Tabs/TabsList.tsx
@@ -42,113 +42,159 @@ export type TabsListProps = React.ComponentPropsWithoutRef<typeof TabsPrimitive.
   alignLeft?: boolean | Breakpoint;
   /** Explicit layout variant. `"hug"` sizes the list to its content; `"fill"` stretches it full-width with equal-width tabs. Takes precedence over `fullWidth`/`alignLeft`. */
   variant?: "fill" | "hug";
+  /**
+   * Drops the leading padding on the first tab so its label sits flush with the
+   * list's left edge, lining it up with page content above or below the bar
+   * (a heading, for instance) rather than being indented by the tab's own
+   * padding. Spacing between tabs is unaffected. Pair with `alignLeft` or
+   * `variant="hug"`. @default false
+   */
+  flushLeft?: boolean;
+  /**
+   * Scrolls the row horizontally when the tabs don't fit, instead of letting
+   * them compress. By default a tab may shrink until its label ellipsises,
+   * which on a narrow viewport can truncate every label at once; here each tab
+   * keeps its natural width and the overflow scrolls. The scrollbar itself is
+   * hidden — the clipped tab at the edge is the affordance.
+   *
+   * Because the tabs are content-sized, they no longer spread to fill the row:
+   * a set that does fit sits left-aligned, as with `alignLeft`.
+   *
+   * Horizontal orientation only: a vertical list already grows downwards.
+   * @default false
+   */
+  scrollable?: boolean;
 };
 
 /** Container for {@link TabsTrigger} elements. Renders a sliding active-tab indicator that animates between tabs. */
 export const TabsList = React.forwardRef<
   React.ComponentRef<typeof TabsPrimitive.List>,
   TabsListProps
->(({ className, children, fullWidth = true, alignLeft, variant, ...props }, ref) => {
-  const innerRef = React.useRef<HTMLDivElement>(null);
-  const indicatorRef = React.useRef<HTMLSpanElement>(null);
+>(
+  (
+    {
+      className,
+      children,
+      fullWidth = true,
+      alignLeft,
+      variant,
+      flushLeft = false,
+      scrollable = false,
+      ...props
+    },
+    ref,
+  ) => {
+    const innerRef = React.useRef<HTMLDivElement>(null);
+    const indicatorRef = React.useRef<HTMLSpanElement>(null);
 
-  React.useImperativeHandle(ref, () => innerRef.current as HTMLDivElement);
+    React.useImperativeHandle(ref, () => innerRef.current as HTMLDivElement);
 
-  const updateIndicator = React.useCallback(() => {
-    const list = innerRef.current;
-    const indicator = indicatorRef.current;
-    if (!list || !indicator) return;
+    const updateIndicator = React.useCallback(() => {
+      const list = innerRef.current;
+      const indicator = indicatorRef.current;
+      if (!list || !indicator) return;
 
-    const activeTab = list.querySelector<HTMLElement>('[data-state="active"]');
-    if (!activeTab) {
-      indicator.style.opacity = "0";
-      return;
-    }
+      const activeTab = list.querySelector<HTMLElement>('[data-state="active"]');
+      if (!activeTab) {
+        indicator.style.opacity = "0";
+        return;
+      }
 
-    const isVertical = list.dataset.orientation === "vertical";
+      const isVertical = list.dataset.orientation === "vertical";
 
-    indicator.style.opacity = "1";
+      indicator.style.opacity = "1";
 
-    if (isVertical) {
-      indicator.style.background =
-        "linear-gradient(180deg, var(--color-buttons-tertiary-default) 0%, var(--color-tab-active) 50%, var(--color-buttons-tertiary-default) 100%)";
-      indicator.style.inset = `0 -1px auto auto`;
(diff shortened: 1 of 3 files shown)


jacksonford-07:
**Storybook:** Components → Tabs → **Scrollable - Narrow Container** and **Full Width - Align Left, Flush**.

Reminder from the description: review the diff with `?w=1`. GitHub reports ~248 changed lines; ignoring whitespace it's 48 insertions / 2 deletions.

src/components/Tabs/TabsList.tsx
```diff
@@ -42,113 +42,159 @@ export type TabsListProps = React.ComponentPropsWithoutRef<typeof TabsPrimitive.
...
+        className={cn(
+          // `isolate` scopes the active indicator's `z-10` to a local stacking
+          // context. Without it, `relative` alone creates no stacking context, so
+          // the indicator escapes and paints above page chrome (e.g. the header).
+          "relative isolate",
+          getLayoutClass(variant, fullWidth, alignLeft),
+          // Horizontal only: in the vertical orientation the leading padding is
+          // the row's left inset, not a label offset, so zeroing it would pull
+          // the first label out of line with the others.
+          flushLeft && "data-[orientation=horizontal]:[&>[role=tab]:first-child]:pl-0",
```
alfonsovo-fv, line 205: I think this misaligns the indicator, which is awkward given it is the thing the prop is for.

The indicator is centred on the tab box and then given the label's width (lines 119 to 120):

```js
indicatorLeft = offsetLeft + offsetWidth / 2 - textWidth / 2
```

With symmetric `px-4` that happens to land exactly on the label. With `pl-0` it does not. `offsetWidth` is `pl + T + pr`, so `indicatorLeft` comes out at `L + (pl + pr) / 2` while the label starts at `L + pl`, and the difference is `(pr - pl) / 2`. The label width cancels out entirely, so it is a flat 8px every time, whatever the tab says.

I measured it against a replica of the trigger's box rather than trust the algebra:

```
default (px-4):   labelLeft 16, indicatorLeft 16, offset 0
flushLeft (pl-0): labelLeft  0, indicatorLeft  8, offset 8
```

So the label lines up with the heading exactly as you intended, and the underline sits 8px to the right of it. Both of your new stories have the first tab active, so Chromatic's diffs should be showing this already, worth a look before anything else.

The fix I would go for is measuring the label directly rather than inferring its position from the box:

```js
const listRect = list.getBoundingClientRect();
const spanRect = (textSpan ?? activeTab).getBoundingClientRect();
const indicatorLeft = spanRect.left - listRect.left + list.scrollLeft;
```

That is immune to any padding change rather than just this one, and the `+ list.scrollLeft` is needed once `scrollable` is in play, because `getBoundingClientRect` is viewport-relative while the transform is in content space.

Worth taking the general shape of this away. The indicator maths had an assumption baked into it, that trigger padding is symmetric, which nothing stated and nothing tested. Adding a prop that quietly breaks a silent assumption is the most common way this class of bug appears, so when you change padding, spacing or sizing, it pays to grep for whatever measures the thing you just changed.

src/components/Tabs/TabsList.tsx
```diff
@@ -42,113 +42,159 @@ export type TabsListProps = React.ComponentPropsWithoutRef<typeof TabsPrimitive.
...
+          // `flex-none` is what stops the tabs compressing — without it they
+          // share the row and never overflow, so there is nothing to scroll.
+          // The scrollbar is hidden in both engines: Firefox reads
+          // `scrollbar-width`, WebKit and Blink the pseudo-element.
+          scrollable &&
+            "data-[orientation=horizontal]:overflow-x-auto data-[orientation=horizontal]:[scrollbar-width:none] data-[orientation=horizontal]:[&::-webkit-scrollbar]:hidden data-[orientation=horizontal]:[&>[role=tab]]:flex-none",
```
alfonsovo-fv, line 211: Does the active tab get scrolled into view on mount? I do not think it does. The layout effect positions the indicator but never touches `scrollLeft`, and Radix only scrolls a tab into view when it takes focus, which mount is not.

So if tab state comes from the URL, someone landing on the fifth tab in a narrow container sees tab one, no indicator anywhere, and nothing suggesting the active tab is off to the right. On the agency insights toolbar that feels likely rather than hypothetical.

Setting `scrollLeft` directly in the same layout effect avoids it, and is safer than `scrollIntoView`, which can scroll ancestor containers and the page with it:

```js
if (scrollable && !isVertical) {
  const overshoot = activeTab.offsetLeft + activeTab.offsetWidth - list.clientWidth;
  if (overshoot > 0) list.scrollLeft = overshoot;
}
```

Two smaller things while I am here, neither blocking.

Hiding the scrollbar means a desktop user with a plain mouse has no way to scroll the row. Trackpad and touch are fine, and keyboard works because arrow keys move focus and focus scrolls, but a wheel does not scroll horizontally by default and there is nothing visible to drag. The clipped tab tells you there is more, it just does not give you a way to reach it. Worth asking design whether that case matters here rather than deciding it in code.

And `scrollable`'s `flex-none` beats `variant="fill"`'s `flex-1` on specificity, so `scrollable` silently wins over `fill`. That is probably the behaviour you want, but the `variant` doc says it takes precedence over the other layout props, so it is now slightly untrue. One clause in the `scrollable` doc covers it.

alfonsovo-fv (changes requested):
Both props are sensible and properly opt-in, and thank you for the `?w=1` note, it checks out exactly: `git diff -w` gives 48 insertions and 2 deletions, everything else is reindentation from the new destructuring wrapper. Telling a reviewer how to read your diff is a genuinely good habit, it changes how carefully this gets looked at.

I also went looking for two problems I expected and did not find them, so for the record: `--shadow-focus-ring` is an *inset* shadow (`theme.css` line 23), so `overflow-x-auto` will not clip the focus rings on the tabs; and `flex-none` beats both `alignLeft`'s `flex-initial` and `fill`'s `flex-1` on specificity rather than on source order, so that override is stable rather than lucky.

Requesting changes for two things though, both visible rather than theoretical.

The first is that `flushLeft` misaligns the active indicator by 8px, which is awkward because it breaks the exact alignment the prop exists to create. The indicator's position is derived from the tab box on the assumption that its padding is symmetric, and `pl-0` breaks that assumption. I measured it rather than just doing the algebra. Details on line 175.

The second is that `scrollable` never scrolls the active tab into view, so a URL-driven tab state can land someone on a page showing tab one with no indicator and no hint that the active tab is off to the right. Details on line 181, along with two smaller notes.

Worth saying that neither of these is a sloppiness thing, they are both the same shape of problem: a new prop breaking an assumption that existing code never stated out loud. That is most of what goes wrong when you add options to a component that measures itself.

src/components/Tabs/TabsList.tsx
jacksonford-07: Fixed in `f1ea543`. Your algebra is exactly right, and I reproduced the 8px in a browser before touching it:

```
flushLeft, padding 0/16:  labelLeft 0, indicatorLeft 8, offset 8
default,   padding 16/16: labelLeft and indicatorLeft agree to 0.3px
```

One deviation from your snippet, and say the word if you would rather have yours. You suggested:

```js
const indicatorLeft = spanRect.left - listRect.left + list.scrollLeft;
```

I went with the label offset inside the tab, added to the tab offset:

```js
const textOffsetInTab = target.getBoundingClientRect().left - activeTab.getBoundingClientRect().left;
const indicatorLeft = activeTab.offsetLeft + textOffsetInTab;
```

Same result. Two small reasons I preferred it. It needs no `scrollLeft` term at all, since `offsetLeft` is already scroll-invariant and the remaining term is a difference of two rects taken in the same frame, so there is no way for the two to disagree mid-scroll. And `offsetLeft` is relative to the padding box, which is also the indicator positioning context, whereas `listRect.left` is the border box, so yours would drift by the border width if the list ever gained one. Neither matters today, the list has no border.

Verified scroll-invariance rather than assuming it: `dL` stays 0.00 at `scrollLeft` 0, 100 and 144 on the scrollable story.

Test pins the formula rather than the outcome, so it fails on the old code with `translateX(8px)` against `translateX(0px)`. It has to stub `getBoundingClientRect` and `offsetLeft`, since jsdom reports zeroes for everything and a naive version of this test would pass either way.

Taking the general point too. The silent assumption was that trigger padding is symmetric, and nothing named or tested it, so the fix is now the thing the test states.

alfonsovo-fv (approved):
Approving. Your formula is better than the one I suggested, and both of your reasons hold: `offsetLeft` is already scroll-invariant so the `scrollLeft` term was redundant in mine, and it is padding-box relative like the indicator's positioning context, where my `listRect.left` was border-box and would drift if the list ever gained a border. Worked it through for both padding cases and at three scroll positions, offset is 0 everywhere. 22 tests pass.

Stubbing `getBoundingClientRect` and `offsetLeft` so the test can actually fail on the old code was the right call too. A version of that test without the stubs would have passed against the bug, which is worse than no test.

One thing to flag rather than hold this up: the scroll-into-view point on my other comment is unaddressed, and I am fine with that landing as a ticket instead. `scrollable` has no consumers yet, so nothing regresses today, but the first URL-driven tab state that uses it will land someone on tab one with no indicator and no hint the active tab is off to the right. Worth raising now while the context is fresh rather than rediscovering it later.

src/components/Tabs/TabsList.tsx
jacksonford-07: You are right, it did not. Fixed in `3c2cebd`, using your `scrollLeft` approach rather than `scrollIntoView` for exactly the reason you gave.

Measuring it turned up something the algebra would not have. Your snippet in the layout effect under-scrolls, because at mount the webfont has not landed yet and the tabs are narrower than they end up. On the new story the row stopped at `scrollLeft` 112 when it needed 145, leaving the active tab 32px outside the container. So the reveal is now a callback that also runs in the existing `fonts.ready` handler, next to the `updateIndicator()` that was already there for the same reason.

Guarded with `overshoot > list.scrollLeft` rather than `> 0`, so it only ever scrolls further right. Otherwise the font-load pass could drag the row back from wherever someone had already scrolled it. Verified: scrolled to 100 by hand, fired a resize, stayed at 100.

Added a story mounted on the third tab, both because it is the regression case and so Chromatic has a snapshot of it. Two tests, and they fail without the fix with 0 against 130; the already-fits case stays green so it is not just always scrolling.

Your two smaller notes I have not actioned, and agree with both. The hidden scrollbar leaving a plain-mouse user no way to scroll is a design question rather than a code one, so I would rather ask than decide it here. And `scrollable`s `flex-none` beating `variant="fill"` does make the `variant` doc untrue as written; happy to add the clause if you want it in this PR rather than a follow-up.

PR #625, opened 2026-07-30
feat(Drawer): add the menu variant

jacksonford-07:
Adds a `menu` variant to `DrawerContent`: an inset floating panel, rounded on every corner and held off all three edges, rather than a sheet flush to the bottom of the viewport.

## Why

It's the mobile form of the Figma `V2 Menu Dropdown`. On a phone a dropdown becomes a bottom sheet, but the V2 menu is drawn as a floating panel with its own margins — not the edge-to-edge sheet `variant="sheet"` produces.

Landing it on its own, ahead of anything that opts in, so the DropdownMenu work that needs it isn't carrying an unrelated Drawer change.

## Blast radius

**Nothing renders differently.** `menu` is a new member of an existing union with no consumers yet — all 12 current `Drawer` call sites keep whatever variant they already pass. `DrawerVariant` is already exported, so `src/index.ts` is untouched and there's no new public surface beyond the union member.

## Verification

- `Drawer` suite — 40 tests passing, including coverage for the variant's insets and radius
- Story added

Labelled `chromatic` even though existing variants are untouched, because the variant map is shared and I'd rather have the baseline than assume.

### DIFF
--- a/src/components/Drawer/Drawer.test.tsx
+++ b/src/components/Drawer/Drawer.test.tsx
@@ -247,8 +247,87 @@ describe("Drawer", () => {
         "inset-x-4",
         "bottom-4",
         "rounded-lg",
+        // Same blur as a sheet: the docs promise the two share the modal surface.
+        "backdrop-blur-[4px]",
       );
-      expect(content).not.toHaveClass("rounded-t-xl");
+      // `rounded-t-xs` is the class that is genuinely in the running: `SLIDE_CLASSES.bottom`
+      // emits it for `position="bottom"`, and `rounded-lg` only wins by being merged
+      // over it. A negative on `rounded-t-xl` would pass whatever happened here, since
+      // only `SHEET_CLASSES` emits that and this is not a sheet.
+      expect(content).not.toHaveClass("rounded-t-xs");
+      expect(content).not.toHaveClass("w-full");
+      expect(content).not.toHaveClass("inset-x-0");
+    });
+
+    it("emits the classes the menu variant overrides, on a plain bottom drawer", async () => {
+      // Proves the negatives above are live rather than decorative: these are the
+      // classes `position="bottom"` really does add, so if the merge ever stopped
+      // stripping them for `menu` the assertions above would start failing.
+      const user = userEvent.setup();
+      render(
+        <Drawer>
+          <DrawerTrigger>Open plain</DrawerTrigger>
+          <DrawerContent position="bottom" aria-describedby={undefined}>
+            <DrawerTitle>Title</DrawerTitle>
+          </DrawerContent>
+        </Drawer>,
+      );
+      await user.click(screen.getByRole("button", { name: "Open plain" }));
+      const content = screen.getByRole("dialog");
+      expect(content).toHaveClass("rounded-t-xs");
+      expect(content).toHaveClass("w-full");
+      expect(content).toHaveClass("inset-x-0");
+    });
+
+    it("lets the variant shadow win instead of the base shadow-lg", async () => {
+      // tailwind-merge does not treat `shadow-blur-menu` as a box-shadow utility, so
+      // it never deduped it against the base `shadow-lg` and `shadow-lg` won on
+      // stylesheet order. Measured in a browser: the menu panel was painting
+      // shadow-lg's `0 4px 8px -1px / 0 8px 22px -1px` rather than the menu token's
+      // `0 6px 12px 0`. `shadow-lg` is now scoped to `panel`.
+      const user = userEvent.setup();
+      const { unmount } = render(
+        <Drawer>
+          <DrawerTrigger>Open menu</DrawerTrigger>
+          <DrawerContent position="bottom" variant="menu" aria-describedby={undefined}>
+            <DrawerTitle>Title</DrawerTitle>
+          </DrawerContent>
+        </Drawer>,
+      );
+      await user.click(screen.getByRole("button", { name: "Open menu" }));
+      expect(screen.getByRole("dialog")).toHaveClass("shadow-blur-menu");
+      expect(screen.getByRole("dialog")).not.toHaveClass("shadow-lg");
+      unmount();
+
+      renderDrawer();
+      await user.click(screen.getByRole("button", { name: "Open drawer" }));
+      expect(screen.getByRole("dialog")).toHaveClass("shadow-lg");
+    });
+
+    it("gives menu and sheet the same blur, since only their shape differs", async () => {
+      const user = userEvent.setup();
+      const { unmount } = render(
+        <Drawer>
+          <DrawerTrigger>Open sheet</DrawerTrigger>
+          <DrawerContent position="bottom" variant="sheet" aria-describedby={undefined}>
+            <DrawerTitle>Title</DrawerTitle>
+          </DrawerContent>
+        </Drawer>,
+      );
+      await user.click(screen.getByRole("button", { name: "Open sheet" }));
+      expect(screen.getByRole("dialog")).toHaveClass("backdrop-blur-[4px]");
+      unmount();
+
+      render(
+        <Drawer>
+          <DrawerTrigger>Open menu</DrawerTrigger>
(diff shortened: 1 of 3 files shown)


jacksonford-07:
**Storybook:** Components → Drawer → **Bottom Menu** (compare against **Bottom Sheet** to see the difference — inset floating panel vs edge-to-edge sheet).

Nothing else renders differently: `menu` has no consumers yet, so all 12 existing Drawer call sites are untouched.

src/components/Drawer/Drawer.tsx
```diff
@@ -28,6 +32,18 @@ export type DrawerVariant = "panel" | "sheet";
...
+ * Surface and inset for the `"menu"` variant. Same modal background and stroke as
+ * a sheet — the design puts this panel on `background/secondary` with
+ * `border/strong`, not on the lighter surface the trigger-anchored dropdown uses.
+ * The 24px radius is the design's `rounded-lg`.
+ *
+ * The insets restate position on purpose — they override `SLIDE_CLASSES.bottom`,
+ * which pins a sheet flush to the bottom edge at full width.
+ */
+const MENU_CLASSES =
+  "inset-x-4 bottom-4 w-auto rounded-lg border border-modal-stroke bg-modal-background shadow-blur-menu backdrop-blur-[8px]";
```
alfonsovo-fv: The blur does not match what the docs promise. `SHEET_CLASSES` is `backdrop-blur-[4px]` and this is `backdrop-blur-[8px]`, but the `DrawerVariant` doc says `menu` "Shares the modal surface with `"sheet"`; only the shape differs", and the comment directly above says "Same modal background and stroke as a sheet".

Blur is part of the surface, so one of the two wants changing: either match sheet at 4px, or keep 8px and add a clause to both doc comments. Worth actually resolving rather than leaving, because the next person adding a variant here will read "only the shape differs" and copy `SHEET_CLASSES` expecting parity.

Separately, and this one is pre-existing rather than something you introduced: the rendered class list carries both `shadow-lg` from the base classes and `shadow-blur-menu` from here. tailwind-merge does not recognise `shadow-blur-menu` as a shadow utility, so it does not dedupe them, which means which shadow actually wins comes down to stylesheet order rather than class order. The `sheet` variant has the same thing and is presumably rendering as intended, so I would not touch it, but worth eyeballing your menu panel in Chromatic to confirm you are getting `shadow-blur-menu` and not `shadow-lg`.

src/components/Drawer/Drawer.test.tsx
```diff
@@ -225,6 +225,33 @@ describe("Drawer", () => {
...
+      await user.click(screen.getByRole("button", { name: "Open menu" }));
+      const content = screen.getByRole("dialog");
+      expect(content).toHaveClass(
+        "bg-modal-background",
+        "border-modal-stroke",
+        "inset-x-4",
+        "bottom-4",
+        "rounded-lg",
+      );
+      expect(content).not.toHaveClass("rounded-t-xl");
```
alfonsovo-fv: This assertion cannot fail. `rounded-t-xl` only ever comes from `SHEET_CLASSES`, which does not apply when `variant` is `menu`, so it was never in the running.

The class that *is* a real candidate is `rounded-t-xs`, which `SLIDE_CLASSES.bottom` adds for `position="bottom"`. Your `rounded-lg` has to beat that through tailwind-merge, and that is the bit worth pinning down:

```ts
expect(content).not.toHaveClass("rounded-t-xs");
```

I dumped the rendered class list to check, and it does work: `rounded-t-xs`, `w-full` and `inset-x-0` are all correctly stripped. So the behaviour is right, the test just is not watching the thing that could actually break.

This is the second one of these I have seen today, the other being `IconButton.test.tsx` line 56, so here is the general version. A negative assertion is only worth writing if the class could plausibly have been there. Ask yourself what would have to go wrong for it to appear, and if the answer is "nothing could put it there", the assertion is decoration. The valuable negatives are the ones naming a class that something else in the chain genuinely emits, because those are the ones that catch a merge or ordering change later.

alfonsovo-fv (approved):
Approving, this is the tidiest one of the batch. Small, genuinely additive, and the description's claims hold up when checked rather than just sounding right.

Things I verified rather than took on trust:

- `--radius-lg` really is 24px (`theme.css` line 29), so the comment is accurate
- The blast radius claim is correct: new union member, no consumers, and `DrawerVariant` was already exported so there is no new public surface
- The override you describe in the `MENU_CLASSES` comment actually works. I rendered the variant and dumped the class list: `SLIDE_CLASSES.bottom` contributes `inset-x-0`, `w-full` and `rounded-t-xs`, and all three are correctly stripped in favour of `inset-x-4 bottom-4 w-auto rounded-lg`. Worth knowing that held, because it is doing real work rather than being cosmetic

Also a note on how you sequenced this. Landing the Drawer variant on its own, ahead of the DropdownMenu work that needs it, is the right instinct: the consumer PR gets to be about the consumer, and if the variant turns out to be wrong it reverts on its own. Same habit as the hook move in 621. Keep doing that.

Two follow-ups inline, neither blocking and both fixable in one push: the blur contradicts what the docs promise, and one test assertion cannot fail.

src/components/Drawer/Drawer.tsx
jacksonford-07: Fixed in `ce7f9e5`, resolved in favour of the docs: menu is now `backdrop-blur-[4px]`.

Checked which side was wrong rather than picking one. `backdrop-blur-[8px]` appears nowhere else in the codebase, while `Dialog` and `sheet` both use `[4px]`, and there is no design reference on the Drawer stories pointing at 8px. So it read as a slip rather than an intentional difference, which makes the docs right.

Your shadow note turned out to be more than an ordering coincidence, and this is the interesting part. It is not that which shadow wins is luck, it is that `shadow-lg` always wins and `shadow-blur-menu` never applies. Measured on the menu panel in a browser:

```
computed box-shadow: 0 4px 8px -1px rgba(0,0,0,0.08), 0 8px 22px -1px rgba(0,0,0,0.1)
--shadow-lg:         0px 4px 8px -1px #00000014, 0px 8px 22px -1px #0000001a
--shadow-blur-menu:  0px 6px 12px 0px #0000001a
```

So both `sheet` and `menu` have been painting `shadow-lg` while naming the menu token. I tried `shadow-[var(--shadow-blur-menu)]` first, hoping tailwind-merge would recognise the arbitrary form as a box-shadow, and it did not dedupe either, presumably classifying it as a shadow colour. So `shadow-lg` is now scoped to `variant === "panel"` instead, which is explicit and does not depend on merge behaviour. The menu panel computes `0 6px 12px 0` and panel is unchanged, both verified in the browser.

One judgement call to flag: that changes `sheet`s rendered shadow too, which is pre-existing rather than introduced here. You said you would not touch it on the basis that it was presumably rendering as intended, and it was not, which I think flips that. It also has to change for the doc contract to hold, since I just fixed the blur for exactly that reason and leaving the shadows different would reintroduce the same contradiction. Happy to pull it out into its own PR if you would rather sheet moved separately. Chromatic will show it either way.

src/components/Drawer/Drawer.test.tsx
jacksonford-07: Fixed in `ce7f9e5`. Now `rounded-t-xs`, plus `w-full` and `inset-x-0`, which are the other two `SLIDE_CLASSES.bottom` emits and the menu has to override.

Added a companion test asserting those three *are* present on a plain bottom drawer, so the negatives cannot quietly become decoration again if the merge changes. That one caught something while I was writing it: my first version used `variant="sheet"` for the positive case and failed, because `SHEET_CLASSES` merges `rounded-t-xl` over `rounded-t-xs`. The plain `panel` variant is the only one that keeps it, which is a neat illustration of your point, the class really is in the running and really does get stripped by whatever comes after it.

Taking the general rule. Both of the ones you found today were the same shape: a negative naming a class nothing in the chain could have emitted. The test to apply is what would have to go wrong for this class to appear, and if the answer is nothing, it is not an assertion.

PR #626, opened 2026-07-30
feat(UserHandle): add a tone prop

jacksonford-07:
Adds a `tone` prop to `UserHandle`, defaulting to the existing muted `secondary`.

## Why

`secondary` assumes the handle sits *beneath* a display name, which carries the emphasis. Figma's V2 table rows identify a row by handle alone — there's no name above it — so the handle needs to be the primary text rather than muted supporting text.

## Blast radius

Defaulted to `secondary`, so the one existing consumer and every future one renders exactly as before. `UserHandleProps` was previously an empty interface extending `HTMLAttributes`, so this is its first real prop.

## Verification

- `UserHandle` suite — 9 tests passing, with coverage for both tones
- `PrimaryTone` story added and `tone` wired into the Storybook controls

### DIFF
--- a/src/components/UserHandle/UserHandle.stories.tsx
+++ b/src/components/UserHandle/UserHandle.stories.tsx
@@ -10,6 +10,7 @@ const meta = {
   tags: ["autodocs"],
   argTypes: {
     children: { control: "text" },
+    tone: { control: "select", options: ["secondary", "primary"] },
   },
   args: {
     children: "fit_aitana",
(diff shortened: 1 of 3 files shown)


jacksonford-07:
Closing — the prop isn't needed.

`cn` uses `twMerge` and applies `className` last, so `<UserHandle className="text-content-primary">` already overrides `text-content-secondary` correctly. That makes `tone` public API for something the existing escape hatch handles.

It also would have been the only `tone` prop in the library, so it introduced a naming convention for a single consumer.

The one real argument for keeping it is constraining the choice: with a raw `className` the next caller passes `text-blue-500` and that's how token drift starts. If the team's rule is that colour must be prop-gated rather than className-overridden, this is worth reopening — happy to be told that's the convention. Until then the leaner option wins, and the agency table passes the token at the call site.

PR #627, opened 2026-07-30
feat(ChartSkeleton): add shaped loading placeholders for charts

jacksonford-07:
Adds `ChartSkeleton` — a wave-animated placeholder shaped like the content it stands in for — and teaches `ChartLoadingOverlay` to use it.

## Why

A centred spinner over a chart tells you nothing about what's arriving, and the layout jumps when the real series lands. A placeholder in the shape of the chart holds the space and reads as "this is a chart, loading".

Six shapes: `area`/`line` (one filled band — a continuous series reads better as a band than as bars), `bar`, `circular` for pie and radial, plus `table` and `rows` for the two list bodies an insight card can hold instead of a chart. `rows` mirrors `Demographics` down to its 8px inner gap and bar height, so the swap to real content is invisible.

## `ChartLoadingOverlay`

Gains `variant?: ChartSkeletonVariant | false`, defaulting to `"area"`. The overlay is now opaque and shows the skeleton; `variant={false}` keeps the old semi-transparent wash with a centred spinner.

**This changes the default rendering** for the 11 existing consumers — they get a shaped skeleton where they used to get a spinner. That's the intended improvement, but it's why this is labelled `chromatic`: worth eyeballing the diffs rather than taking my word for it. Any consumer that prefers the spinner can pass `variant={false}`.

## Verification

- Chart suite — 103 tests passing, including per-variant coverage and a11y for the two list variants
- `pnpm build` clean. (Dropped the previous `size-limit` line: it measures `dist/index.mjs` and this ships in `charts.mjs`, so it could not have failed for this PR and told us nothing.)
- Stories added: `ChartSkeletons` (all six shapes) and `LoadingOverlayWithSkeleton`

## Note on the stack

First of three carving up the chart work, so no single PR mixes additive components with visual rewrites. The next two — `ChartCard`'s slots and loading header, then the tooltip's menu surface — are stacked on this one because `charts.ts`, `Chart.test.tsx` and `ChartHelpers.stories.tsx` are shared and would otherwise conflict repeatedly. Only the skeleton's own slice of those three files is in here.

## Review follow-ups

**Loading is now announced.** Swapping the overlay default from `Loader` to `ChartSkeleton` had silently removed the only signal telling assistive tech the chart was loading, for all 11 consumers, because every `Skeleton` is `aria-hidden`. Rendered both paths to confirm rather than reason about it:

| | `role=status` | `<output>` | accessible "loading" |
| --- | --- | --- | --- |
| skeleton (before) | 0 | 0 | 0 |
| spinner | 1 | 1 | 1 |

The overlay now carries `role="status"`, `aria-busy="true"` and an overridable `loadingLabel` (default `"Loading chart"`), following `Loader`'s own overridable `ariaLabel` rather than hardcoding English. Scoped to the skeleton path so the spinner path does not nest a live region inside `Loader`'s `<output>`.

The axe assertion that used to cover this could never have failed — axe reports violations in markup that exists, and `aria-hidden` markup has nothing to violate. Replaced with a direct `toHaveAccessibleName(/loading/i)`, plus coverage for a translated label and for not announcing when `loading={false}`.

**The donut hole is masked, not painted.** It was an opaque `bg-background-primary` disc, which only disappears on a surface of exactly that colour — and this renders standalone via `ChartCard`'s `skeleton` prop, not just inside the overlay. Now a radial mask, so the hole is a hole.

**`rows` drives the bar count.** A three-bar chart was getting a seven-bar skeleton, which is the shift the prop exists to prevent. Resolved per variant (`rows ?? 7` for bars, `rows ?? 5` for lists) so seven stays the bar default rather than becoming the only option, and no existing consumer's bar count moves.

## The default is deliberately not `variant={false}`

Calling this out explicitly rather than leaving Chromatic to reveal it, since it is the one thing here a consumer cannot discover from the types.

**This PR changes what the 11 existing `ChartLoadingOverlay` consumers render**: a shaped skeleton over an opaque `bg-background-primary`, where they previously got a centred spinner over a `bg-surface-primary/60` wash. The new behaviour is better and the loading state is now announced in both paths, so the default stays. Any consumer that wants the old rendering passes `variant={false}`.

Two consequences worth knowing. The overlay is opaque and specific rather than a tint, so a chart sitting on some other surface gets a matched block rather than a wash of whatever is beneath it — worth a look in the Chromatic diffs. And this ships on a minor, so it wants a release-note line.

### DIFF
--- a/src/components/Chart/ChartLoadingOverlay.tsx
+++ b/src/components/Chart/ChartLoadingOverlay.tsx
@@ -1,23 +1,34 @@
 import * as React from "react";
 import { cn } from "../../utils/cn";
 import { Loader } from "../Loader/Loader";
+import { ChartSkeleton, type ChartSkeletonVariant } from "./ChartSkeleton";
 
 /** Props for {@link ChartLoadingOverlay}. */
 export interface ChartLoadingOverlayProps extends React.HTMLAttributes<HTMLDivElement> {
   /** Whether to show the loading overlay. @default false */
   loading?: boolean;
+  /**
+   * Shape of the wave-animated {@link ChartSkeleton} shown while loading. Set it
+   * to match the chart underneath — `circular` for pie/radial, `bar` for bar
+   * charts. Pass `false` to fall back to the legacy centred spinner.
+   * @default "area"
+   */
+  variant?: ChartSkeletonVariant | false;
   /** Chart content to render underneath the overlay. */
   children: React.ReactNode;
 }
 
 /**
- * A positioned overlay that displays a loading spinner on top of chart content.
- * The children are always rendered to maintain layout dimensions; the overlay
- * covers them with a semi-transparent background and a centered spinner.
+ * A positioned overlay shown on top of chart content while it loads. The children
+ * are always rendered to maintain layout dimensions.
+ *
+ * By default the overlay is opaque and shows a wave-animated
+ * {@link ChartSkeleton} shaped like the chart. At `variant={false}` it falls back
+ * to a semi-transparent wash with a centred spinner.
  *
  * @example
  * ```tsx
- * <ChartLoadingOverlay loading={isFetching}>
+ * <ChartLoadingOverlay loading={isFetching} variant="area">
  *   <ChartContainer config={config} className="min-h-48">
  *     <LineChart data={data}>...</LineChart>
  *   </ChartContainer>
@@ -25,13 +36,18 @@ export interface ChartLoadingOverlayProps extends React.HTMLAttributes<HTMLDivEl
  * ```
  */
 export const ChartLoadingOverlay = React.forwardRef<HTMLDivElement, ChartLoadingOverlayProps>(
-  ({ loading = false, children, className, ...props }, ref) => {
+  ({ loading = false, variant = "area", children, className, ...props }, ref) => {
     return (
       <div ref={ref} className={cn("relative", className)} {...props}>
         {children}
         {loading && (
-          <div className="absolute inset-0 z-10 flex items-center justify-center bg-surface-primary/60">
-            <Loader show center />
+          <div
+            className={cn(
+              "absolute inset-0 z-10 flex items-center justify-center",
+              variant ? "bg-background-primary" : "bg-surface-primary/60",
+            )}
+          >
+            {variant ? <ChartSkeleton variant={variant} /> : <Loader show center />}
           </div>
         )}
       </div>
--- a/src/components/Chart/ChartLoadingOverlay.tsx
+++ b/src/components/Chart/ChartLoadingOverlay.tsx
@@ -14,6 +14,12 @@ export interface ChartLoadingOverlayProps extends React.HTMLAttributes<HTMLDivEl
    * @default "area"
    */
   variant?: ChartSkeletonVariant | false;
+  /**
+   * Accessible name for the loading region, announced while `loading` is true.
+   * Pass a translated string for i18n. Only applies to the skeleton path; the
+   * `variant={false}` spinner carries its own label. @default "Loading chart"
+   */
+  loadingLabel?: string;
   /** Chart content to render underneath the overlay. */
   children: React.ReactNode;
 }
@@ -36,7 +42,17 @@ export interface ChartLoadingOverlayProps extends React.HTMLAttributes<HTMLDivEl
  * ```
  */
 export const ChartLoadingOverlay = React.forwardRef<HTMLDivElement, ChartLoadingOverlayProps>(
-  ({ loading = false, variant = "area", children, className, ...props }, ref) => {
+  (
+    {
+      loading = false,
+      variant = "area",
+      loadingLabel = "Loading chart",
+      children,
+      className,
+      ...props
+    },
+    ref,
+  ) => {
     return (
       <div ref={ref} className={cn("relative", className)} {...props}>
         {children}
@@ -46,6 +62,17 @@ export const ChartLoadingOverlay = React.forwardRef<HTMLDivElement, ChartLoading
               "absolute inset-0 z-10 flex items-center justify-center",
               variant ? "bg-background-primary" : "bg-surface-primary/60",
             )}
+            // Every `Skeleton` is `aria-hidden`, so without this the skeleton path
+            // announces nothing where the spinner announced through its `<output>`.
+            // This is the contract `Skeleton`'s own doc comment asks callers to honour.
+            // Scoped to the skeleton path deliberately: adding it to the spinner path
+            // would nest a live region inside a live region and risk a double
+            // announcement.
+            {...(variant && {
+              role: "status",
+              "aria-busy": true,
+              "aria-label": loadingLabel,
+            })}
           >
             {variant ? <ChartSkeleton variant={variant} /> : <Loader show center />}
           </div>
--- a/src/components/Chart/ChartSkeleton.tsx
+++ b/src/components/Chart/ChartSkeleton.tsx
@@ -1,0 +1,147 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+import { Skeleton } from "../Skeleton/Skeleton";
+
+/**
+ * Shape the skeleton should imitate while data loads. `area` and `line` render
+ * the same filled band — a continuous series reads better as a band than as
+ * discrete bars. `table` and `rows` stand in for the two list bodies an insight
+ * card can hold rather than a chart: a ranked table, and a breakdown of labelled
+ * bars.
+ */
+export type ChartSkeletonVariant = "area" | "line" | "bar" | "circular" | "table" | "rows";
+
+/** Props for {@link ChartSkeleton}. */
+export interface ChartSkeletonProps extends React.HTMLAttributes<HTMLDivElement> {
+  /** Which shape to imitate. @default "area" */
+  variant?: ChartSkeletonVariant;
+  /**
+   * How many rows to draw, for the `table` and `rows` variants. Match the row
+   * count the loaded card will show, so the card does not resize once the real
+   * content arrives. @default 5
+   */
+  rows?: number;
+}
+
+const BAR_HEIGHTS = ["45%", "70%", "55%", "85%", "60%", "95%", "75%"];
+
+/** Varying label widths, so the rows do not read as a single solid block. */
+const LABEL_WIDTHS = ["w-24", "w-32", "w-20", "w-28", "w-24"];
+
+const labelWidth = (index: number) => LABEL_WIDTHS[index % LABEL_WIDTHS.length];
+
+const Bar = ({ className }: { className?: string }) => (
+  <Skeleton animation="wave" variant="rounded" className={className} />
+);
+
+/**
+ * A ranked table: the header, then one 48px row per entry holding a rank, an
+ * avatar and handle, and a value at the end — the columns {@link ChartCard}'s
+ * table bodies use.
+ */
+const TableSkeleton = ({ rows }: { rows: number }) => (
+  <div className="flex w-full flex-col">
+    <div className="flex h-8 items-center gap-3">
+      <Bar className="h-3 w-4" />
+      <Bar className="h-3 w-16" />
+      <Bar className="ml-auto h-3 w-12" />
+    </div>
+    {Array.from({ length: rows }, (_, index) => (
+      <div key={`table-row-${index}`} className="flex h-12 items-center gap-3">
+        <Bar className="h-4 w-4" />
+        <Skeleton animation="wave" variant="circular" className="size-6 shrink-0" />
+        <Bar className={cn("h-4", labelWidth(index))} />
+        <Bar className="ml-auto h-4 w-16" />
+      </div>
+    ))}
+  </div>
+);
+
+/**
+ * A breakdown list: per row a leading glyph and label with its share at the far
+ * end, then the bar underneath. Mirrors `Demographics`, down to its 8px inner gap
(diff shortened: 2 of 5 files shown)


jacksonford-07:
**Storybook:** Components → Charts → Helpers → **Chart skeletons (per chart type)** and **Loading Overlay With Skeleton**.

I've just pushed a fix worth knowing about: the skeletons story hardcoded four variants, so `table` and `rows` shipped with no story coverage. It now renders all six, with the two list shapes in an auto-height container since forcing the chart height on them misrepresents how they size.

Chromatic re-triggered after that push (a push alone doesn't re-run it). The diffs to actually read are the **11 existing `ChartLoadingOverlay` consumers** — their default changes from a centred spinner to a shaped skeleton.

src/components/Chart/ChartLoadingOverlay.tsx
```diff
@@ -1,37 +1,53 @@
...
         {loading && (
-          <div className="absolute inset-0 z-10 flex items-center justify-center bg-surface-primary/60">
-            <Loader show center />
+          <div
+            className={cn(
+              "absolute inset-0 z-10 flex items-center justify-center",
+              variant ? "bg-background-primary" : "bg-surface-primary/60",
+            )}
+          >
+            {variant ? <ChartSkeleton variant={variant} /> : <Loader show center />}
```
alfonsovo-fv, line 77: This is the blocker. Swapping the default from `Loader` to `ChartSkeleton` silently removes the only thing telling a screen reader the chart is loading, for all 11 consumers.

`Loader` renders `<output aria-label="loading">`, and `output` has an implicit `role="status"`, so it announces. Every `Skeleton` sets `aria-hidden="true"` (`Skeleton.tsx` line 69), so the skeleton announces nothing, and neither this overlay nor `ChartSkeleton` sets `aria-busy` or `role="status"`. I rendered both paths to check:

```
new default:      role=status/output NO,  accessible "loading" label 0
variant={false}:  role=status/output YES, accessible "loading" label 1
```

So a sighted user gets a better loading state and a screen reader user gets nothing at all.

`Skeleton`'s own doc comment already spells out the fix, line 40: "When used as a loading state, wrap the skeleton region with `aria-busy=\"true\"`". So something like:

```tsx
<div role="status" aria-busy="true" aria-label="Loading chart" className={...}>
```

That keeps the announcement in both paths and honours the contract the DS already documents.

The bit I would really take away though: your axe test passes here and could never fail. axe inspects the markup that exists and reports violations in it, and an `aria-hidden` div has nothing to violate. It has no way to tell you something is *missing*. Automated a11y checks are a floor, not a verdict, so a claim like "loading is announced" wants asserting directly:

```ts
expect(screen.getByRole("status")).toHaveAccessibleName(/loading/i);
```

General habit: any time you change what renders in a given state, ask what the old thing was telling assistive tech that the new one is not.

src/components/Chart/ChartLoadingOverlay.tsx
```diff
@@ -1,37 +1,53 @@
...
     return (
       <div ref={ref} className={cn("relative", className)} {...props}>
         {children}
         {loading && (
-          <div className="absolute inset-0 z-10 flex items-center justify-center bg-surface-primary/60">
-            <Loader show center />
+          <div
+            className={cn(
+              "absolute inset-0 z-10 flex items-center justify-center",
+              variant ? "bg-background-primary" : "bg-surface-primary/60",
```
alfonsovo-fv, line 63: Separate from the a11y point: this is a bigger visual change than the skeleton on its own. The old `bg-surface-primary/60` was a wash that blended over whatever the chart sat on. `bg-background-primary` is opaque and specific, so any consumer whose chart is not on that surface gets a mismatched block rather than a tint.

Worth going through Chromatic's diffs for the 11 consumers carefully, since that is exactly the kind of thing that shows up there and nowhere else.

A broader question worth answering in the description rather than in code. Should the default have been `variant={false}`? Everything else in your batch has been scrupulously additive, and this one deliberately changes what existing consumers render, on a minor version. I do not think that is the wrong call, the new behaviour is better and you flagged it honestly. But it is the one decision here that a consumer cannot discover without reading the release notes, so if it stays as-is it wants calling out there explicitly rather than left to Chromatic to reveal.

src/components/Chart/ChartSkeleton.tsx
```diff
@@ -0,0 +1,147 @@
...
+    if (variant === "circular") {
+      return (
+        <div
+          ref={ref}
+          className={cn("flex size-full items-center justify-center", className)}
+          {...props}
+        >
+          <div className="relative aspect-square h-full">
+            <Skeleton animation="wave" variant="circular" className="size-full" />
+            <div className="absolute inset-[27%] rounded-full bg-background-primary" />
```
alfonsovo-fv: This works inside `ChartLoadingOverlay` because the overlay paints the same `bg-background-primary`, but it will not hold up standalone, and the doc comment above explicitly invites standalone use via `ChartCard`'s `skeleton` prop. Punch an opaque disc over a card whose surface is anything else and you get a visible wrong-coloured circle in the middle of the donut.

A radial mask avoids caring what is underneath:

```tsx
className="size-full [mask-image:radial-gradient(circle,transparent_27%,black_28%)]"
```

on the `Skeleton`, dropping the absolutely positioned child entirely. Then the hole is genuinely a hole rather than a patch that happens to match.

Same class of thing as the `border-box` detail you spotted in the Chip PR: it works for the arrangement you tested, while quietly assuming something about its surroundings. Worth asking of anything that paints a colour to fake a gap.

src/components/Chart/ChartSkeleton.tsx
```diff
@@ -0,0 +1,147 @@
...
+  variant?: ChartSkeletonVariant;
+  /**
+   * How many rows to draw, for the `table` and `rows` variants. Match the row
+   * count the loaded card will show, so the card does not resize once the real
+   * content arrives. @default 5
+   */
+  rows?: number;
+}
+
+const BAR_HEIGHTS = ["45%", "70%", "55%", "85%", "60%", "95%", "75%"];
```
alfonsovo-fv, line 27: Small one. `BAR_HEIGHTS` is a fixed seven, and `rows` only drives `table` and `rows`, so a bar chart with three bars gets a seven-bar skeleton and the layout shifts on load. That is the exact thing your `rows` doc says the prop exists to prevent, so it seems worth driving it the same way:

```ts
const barHeight = (index: number) => BAR_HEIGHTS[index % BAR_HEIGHTS.length];
```

then map over `rows` and index into it, exactly as `labelWidth` already does. Seven becomes the default rather than the only option, and the prop means the same thing across every variant that uses it.

alfonsovo-fv (changes requested):
The component itself is good. Six sensible shapes, `forwardRef` and `displayName` both there, per-variant tests, and correctly exported from `charts.ts` rather than `index.ts` since charts is its own entry point, so the template's `src/index.ts` item genuinely does not apply. The stacked-PR reasoning is sound too, and splitting the additive component away from the visual rewrites is the right way round.

Requesting changes for one thing: swapping the overlay default from the spinner to the skeleton silently removes the only signal that told assistive tech the chart was loading, and it does that for all 11 consumers at once.

`Loader` renders `<output aria-label="loading">`, and `output` carries an implicit `role="status"`, so it announces. Every `Skeleton` sets `aria-hidden="true"` (`Skeleton.tsx` line 69), and neither this overlay nor `ChartSkeleton` sets `aria-busy` or `role="status"`. I rendered both paths rather than assume:

```
new default:      role=status or <output> NO,  accessible "loading" label 0
variant={false}:  role=status or <output> YES, accessible "loading" label 1
```

So the change makes the loading state better for sighted users and non-existent for screen reader users. `Skeleton`'s own doc comment already states the contract that fixes it. Details and a suggested fix on line 50.

Three smaller things inline: the donut hole assumes what is underneath it, the bar count is fixed where `rows` is configurable, and the opaque background is a bigger visual change than the skeleton alone.

One accuracy note on the description while I am here. `size-limit` being unchanged at 193.14 kB is expected rather than reassuring: it only measures `dist/index.mjs`, and this ships in `charts.mjs`, so that line does not actually tell us anything about this PR's weight. Worth either measuring the right bundle or dropping the claim, because a verification list is only useful if every line on it could have failed.

src/components/Chart/ChartLoadingOverlay.tsx
jacksonford-07: Fixed in `fd7c6aa`. You were right, and I rendered both paths myself before changing anything rather than taking the reasoning on trust:

```
skeleton: role=status 0, <output> 0, accessible "loading" 0, aria-busy 0
spinner:  role=status 1, <output> 1, accessible "loading" 1
```

The overlay now carries `role="status"` and `aria-busy="true"`, which is the contract `Skeleton`s own doc asks callers to honour.

Two deviations from your snippet, both deliberate.

The label is a prop, `loadingLabel`, defaulting to `"Loading chart"`. `Loader` already takes an overridable `ariaLabel`, so a hardcoded English string here would be the odd one out, and you have flagged unforwardable a11y labels before on `UserItem`.

And the ARIA is scoped to the skeleton path rather than applied to both. On the spinner path the wrapper would sit outside `Loader`s `<output>`, nesting a live region inside a live region, which can announce twice. The spinner path already announces on its own, so it needs nothing.

Your point about the test is the one I have taken most from this. Replaced with `expect(screen.getByRole("status")).toHaveAccessibleName(/loading/i)`, plus one for the translated label and one asserting nothing announces when `loading={false}`. The axe check stays, but it is no longer carrying a claim it cannot support.

src/components/Chart/ChartSkeleton.tsx
jacksonford-07: Fixed in `fd7c6aa`, using the radial mask. The absolutely positioned child is gone entirely, so the hole is a hole rather than a patch that happens to match.

Checked the mask is not silently dropped rather than assuming Tailwind emitted something usable: `CSS.supports("mask-image", "radial-gradient(circle, transparent 27%, black 28%)")` is true, and the computed value resolves to `radial-gradient(circle, rgba(0, 0, 0, 0) 27%, rgb(0, 0, 0) 28%)`. Zero `bg-background-primary` discs left in the render.

One honest gap in the verification: I could not capture a screenshot of it. The tool timed out repeatedly on that page, I assume because the wave animations keep the renderer busy. So the evidence here is the computed mask rather than an image, and Chromatic is the real check.

Taking the general shape of it too, and it is the more useful half. Painting a colour to fake a gap encodes an assumption about the surroundings that nothing states and nothing tests. Same as the `border-box` thing on the Chip PR.

src/components/Chart/ChartSkeleton.tsx
jacksonford-07: Fixed in `fd7c6aa`. `rows` now drives the bar count, indexing into `BAR_HEIGHTS` with the modulo exactly as `labelWidth` already did.

One change from your snippet. Driving it from `rows` directly would have moved every existing bar chart from seven bars to the list default of five, which is a visual change nobody asked for. So the default is resolved per variant, `rows ?? 7` for bars against `rows ?? 5` for lists, which I think is what you meant by seven becoming the default rather than the only option. `rows` is now undefaulted in the signature and resolved inside.

Two tests: `rows={3}` gives three bars, and `rows={9}` gives nine with the heights cycling rather than running out.

src/components/Chart/ChartLoadingOverlay.tsx
jacksonford-07: Answered in the description rather than in code, since this was the decision question rather than a defect.

The default stays the skeleton. Added a section spelling out that this changes what the 11 existing consumers render, a shaped skeleton over an opaque `bg-background-primary` where they had a spinner over a `bg-surface-primary/60` wash, that `variant={false}` restores the old rendering, and that it wants a release-note line since it ships on a minor. You are right that it was the one thing here a consumer could not discover from the types, and leaving Chromatic to reveal it was not good enough.

On the opaque background itself I have left it, since the overlay does have to hide the chart underneath and a wash cannot. Worth noting the mask fix on the donut removes the part that actually worried me: the skeleton no longer depends on the overlay painting a particular colour, so the two are no longer coupled by an unstated assumption. What remains is just that the overlay is opaque, which is deliberate.

Also took your accuracy note on the verification list. Dropped the `size-limit` line rather than leaving it: it measures `dist/index.mjs` and this ships in `charts.mjs`, so it could not have failed for this PR. Fair hit, and it is the second thing on that list I would have skimmed past.

alfonsovo-fv (approved):
Approving. Checked all four rather than trusting the diff:

```
skeleton path    role=status, name "Loading chart", aria-busy true
spinner path     exactly 1 status region, so no nested live region
not loading      no status region at all
donut            0 painted discs, 1 masked
bars             default 7, rows=3 gives 3, rows=9 gives 9 with heights cycling
```

63 tests pass. All three of your deviations from my suggestions were improvements, and worth saying why each one was right.

`loadingLabel` as a prop rather than my hardcoded string: correct, and consistent with `Loader`'s `ariaLabel`. An unforwardable English string in a library component is a bug we would have found later.

Scoping the ARIA to the skeleton path: also correct, and I had not thought about it. My version would have wrapped `Loader`'s `<output>` in a second live region. I confirmed the spinner path still has exactly one.

`rows ?? 7` for bars against `rows ?? 5` for lists: correct, and my snippet was wrong. Driving bars from `rows` directly would have moved every existing bar chart from seven bars to five, a silent visual change to fix a layout-shift bug. Good catch on my suggestion.

And thank you for putting the default-change section in the description and dropping the `size-limit` line rather than defending it. The release-note point was the whole ask.

PR #628, opened 2026-07-30
feat(TooltipLabel): add the underlined label trigger and two-row tooltip

jacksonford-07:
Adds `TooltipLabel` — a text label that is itself the tooltip trigger, marked by a dashed underline — plus the `title` prop on `TooltipContent` that it needs.

## Why

The pattern this replaces is a label beside a separate info glyph. That has two problems: the glyph is an icon button, so it stretches the row past text height, and the accessible name ends up on the glyph rather than the words the user is reading.

Here the visible text *is* the trigger, so it carries its own accessible name and the row stays at text height. The label is repeated as the tooltip's heading, giving the two-row block the design specifies — the term, then what it means.

## `TooltipContent.title`

Optional heading above the body: bold label, plain-weight description. Without it the tooltip is a single bold line exactly as before, so **all 62 existing consumers are unchanged**.

One type detail: `TooltipContentProps` now `Omit`s `title` from the underlying HTML attributes, because `HTMLAttributes` already declares it as `string` and we need `ReactNode`. That's a narrowing on a prop nobody was passing as a native tooltip attribute.

## Touch behaviour

Radix tooltips don't open on touch — its trigger ignores pointer events whose `pointerType` is `"touch"`, so focus is the only thing that opens them, and the trigger closes on click. A tap therefore made the tooltip flash and vanish. Worse, once the label had focus, dismissing and tapping again fired no focus event at all, so it couldn't be reopened.

`TooltipLabel` controls its own open state on touch and toggles from the tap, snapshotting the state at `pointerdown` because Radix's own close lands between `pointerdown` and `click`. Dismissal still comes from Radix — outside tap, Escape, scroll, another tooltip opening. Mouse and keyboard paths are untouched.

## Verification

- Tooltip suite — 25 tests passing, including the tap-to-open, tap-to-reopen-after-dismiss and toggle-closed sequences, and one asserting a **mouse** click still dismisses, which is what proves the change is scoped to touch
- `pnpm build` clean, `TooltipLabel` present in `dist/index.d.ts`
- Story added

## Deliberately not in here

`Tooltip.tsx` on the insights branch also carries a 700ms → 200ms provider-delay default and fade animations. Both are left out: the delay change affects every tooltip in the product and needs to stay independently revertable, especially since `ChartCard` will depend on `TooltipLabel` landing.

### DIFF
--- a/src/components/Tooltip/TooltipLabel.tsx
+++ b/src/components/Tooltip/TooltipLabel.tsx
@@ -1,0 +1,118 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "./Tooltip";
+
+/** Props for {@link TooltipLabel}. */
+export interface TooltipLabelProps extends React.ComponentPropsWithoutRef<"button"> {
+  /** Tooltip content. Pass a translated string for i18n. */
+  tooltip: React.ReactNode;
+  /** The label text, which doubles as the trigger. */
+  children: React.ReactNode;
+}
+
+/**
+ * A text label that is itself the tooltip trigger, marked by a dashed
+ * underline. Use this instead of pairing a label with a separate info glyph:
+ * the visible text becomes the trigger's accessible name, and the row stays at
+ * text height rather than being stretched by an icon button.
+ *
+ * The label is repeated as the tooltip's heading, so the tooltip reads as the
+ * two-row block the design specifies — the term, then what it means.
+ *
+ * Tapping the label toggles the tooltip on touch devices, where there is no
+ * hover to trigger it. Tapping elsewhere, scrolling, or opening another tooltip
+ * dismisses it.
+ *
+ * @example
+ * ```tsx
+ * <TooltipLabel tooltip="Total earnings for the selected period.">Total Earnings</TooltipLabel>
+ * ```
+ */
+export const TooltipLabel = React.forwardRef<HTMLButtonElement, TooltipLabelProps>(
+  ({ tooltip, children, className, onClick, onPointerDown, onBlur, ...props }, ref) => {
+    /*
+     * Touch has to drive the tooltip from the tap itself, not from focus.
+     *
+     * Radix's trigger ignores pointer events whose `pointerType` is "touch", so
+     * hover never opens the tooltip on a phone; focus is the only thing that
+     * does. That is not enough on its own for two reasons, and both need the
+     * open state controlled here:
+     *
+     *   - The trigger closes on click. Since the tap that focuses the label is
+     *     also a click, focus opens the tooltip and the click immediately shuts
+     *     it, so it only flashes.
+     *   - After the first tap the label keeps focus. Dismissing the tooltip by
+     *     tapping elsewhere leaves it focused, so a second tap fires no focus
+     *     event and nothing would reopen it.
+     *
+     * Radix composes trigger handlers with `checkForDefaultPrevented`, so
+     * calling `preventDefault` makes it skip its own open/close and leaves the
+     * toggle below in charge. Dismissal still comes from Radix: `TooltipContent`
+     * closes on outside pointerdown, Escape, scroll, and when another tooltip
+     * opens.
+     *
+     * All of this is scoped to touch, so mouse and keyboard behave as before.
+     */
+    const [open, setOpen] = React.useState(false);
+    const pointerTypeRef = React.useRef<string>("");
+    /**
+     * The open state as it was when the tap began. The trigger's own pointerdown
+     * close lands between pointerdown and click, so toggling against the live
+     * state would read the value Radix just reset and reopen on every tap.
+     */
+    const openAtTapStartRef = React.useRef(false);
+
+    const isTouch = () => pointerTypeRef.current === "touch";
+
+    return (
+      <TooltipProvider>
+        <Tooltip open={open} onOpenChange={setOpen}>
+          <TooltipTrigger asChild>
+            <button
+              ref={ref}
+              type="button"
+              className={cn(
+                "typography-body-small-14px-regular cursor-help rounded-2xs text-left text-content-secondary outline-none",
+                // Anchor the 4-on/4-off tile to the left edge. A centred position (the
+                // `bg-bottom` default) lays the pattern out from the middle outward and
+                // clips mid-dash at the start of the label as well as the end.
+                //
+                // No bottom padding: the dash sits in the descender space the 18px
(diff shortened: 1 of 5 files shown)


jacksonford-07:
**Storybook:** Components → Tooltip → **Label as trigger (dashed underline)**.

Two things a static view won't show, both covered by tests: tapping the label opens the tooltip on touch (Radix tooltips don't open on touch at all by default), and it reopens after being dismissed — which the first attempt at this got wrong, because Radix's focus-open only fires once.

Not in this PR: the 700ms → 200ms provider delay and the fade animations, both left out so the delay stays independently revertable.

src/components/Tooltip/TooltipLabel.tsx
```diff
@@ -0,0 +1,118 @@
...
+                // both a touchscreen and a keyboard.
+                pointerTypeRef.current = "";
+                onBlur?.(event);
+              }}
+              {...props}
+            >
+              {children}
+            </button>
+          </TooltipTrigger>
+          <TooltipContent title={children}>{tooltip}</TooltipContent>
```
alfonsovo-fv: Right block visually, but it announces the label twice. Radix wires the whole content as the trigger's `aria-describedby`, so I resolved it:

```
name:         "Total Earnings"
described by: "Total EarningsTotal earnings for the selected period."
```

The fix belongs here rather than in `TooltipContent`, since the duplication is specific to this component (a `title` on an icon-button trigger does need announcing):

```tsx
<TooltipContent title={<span aria-hidden>{children}</span>}>{tooltip}</TooltipContent>
```

Worth knowing your axe test passes on this and always would. Duplication is not a violation, so nothing automated will flag it.

src/components/Tooltip/TooltipLabel.tsx
```diff
@@ -0,0 +1,118 @@
...
+     * The open state as it was when the tap began. The trigger's own pointerdown
+     * close lands between pointerdown and click, so toggling against the live
+     * state would read the value Radix just reset and reopen on every tap.
+     */
+    const openAtTapStartRef = React.useRef(false);
+
+    const isTouch = () => pointerTypeRef.current === "touch";
+
+    return (
+      <TooltipProvider>
```
alfonsovo-fv, line 71: Heads up rather than a change request, since `ChartCard` line 114 and `SwitchField` line 97 already do the same thing.

Radix resolves `delayDuration` from the nearest provider (`delayDurationProp ?? providerContext.delayDuration`, 700ms default), so a bare nested one overrides whatever the app set. Your description mentions a 700ms to 200ms provider default coming on the insights branch, and it will not reach this component, `ChartCard`, or `SwitchField`. Worth a ticket covering all three rather than anything here.

src/components/Tooltip/TooltipLabel.tsx
```diff
@@ -0,0 +1,118 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "./Tooltip";
+
+/** Props for {@link TooltipLabel}. */
+export interface TooltipLabelProps extends React.ComponentPropsWithoutRef<"button"> {
```
alfonsovo-fv: This extends button props, so the native `title` attribute comes along and a consumer could end up with a browser tooltip fighting the Radix one on the same element.

You already `Omit` `title` on `TooltipContentProps` for a type reason. Same prop, worth the same here for a behaviour one:

```ts
extends Omit<React.ComponentPropsWithoutRef<"button">, "title">
```

alfonsovo-fv (approved):
Approving. I checked the touch workaround properly because the claim is an unusual one, and it holds up: Slot merges handlers child-first with no `defaultPrevented` check (`react-slot/dist/index.mjs` 83-85), but Radix's own handler is `composeEventHandlers(props.onClick, context.onClose)` (`react-tooltip/dist/index.mjs` 206), which is guarded by `!event.defaultPrevented`. So the `preventDefault` genuinely does suppress Radix's close, and the `pointerdown` snapshot is doing real work rather than being belt-and-braces. Tests pass, and the 25 count checks out (18 unit plus 7 storybook stories).

Three small things inline, none blocking.

src/components/Tooltip/TooltipLabel.tsx
jacksonford-07: Fixed in `6d209aa`. Reproduced your numbers first, then took the `aria-hidden` heading exactly as you suggested.

A detail worth passing on, because it nearly fooled me. My first attempt at measuring this read `aria-describedby` and joined the `textContent` of the referenced nodes, which reported the duplication as *still there* after the fix. `textContent` ignores `aria-hidden` completely, so that probe could not have detected the fix either way. Switching to the real computed value showed it:

```
name:        "Total Earnings"
description: "Total earnings for the selected period."
```

So the test asserts `toHaveAccessibleDescription` rather than markup, and it does fail against the exposed heading. Same trap as your axe point, one level down: I had swapped an automated check that could not fail for a hand-rolled one that also could not fail.

Second test pins that the heading is still visible, since hiding it from AT and hiding it from the screen are one character apart and only one of them is wanted.

src/components/Tooltip/TooltipLabel.tsx
jacksonford-07: Done in `6d209aa`: `extends Omit<React.ComponentPropsWithoutRef<"button">, "title">`. Agreed that the same prop wants the same treatment for a behaviour reason as for a type one.

src/components/Tooltip/TooltipLabel.tsx
jacksonford-07: Good catch, and agreed it wants covering across all three rather than here. Not changing the nesting in this PR: pulling the provider out would change where consumers have to mount it, which is a bigger call than this PR should make quietly.

Confirmed the mechanism before agreeing: Radix resolves `delayDurationProp ?? providerContext.delayDuration` against the nearest provider, so the app-level value never reaches a component that mounts its own. Same in `ChartCard` and `SwitchField`.

Raising a ticket for the three of them and will link it here.

jacksonford-07: #633, 200ms.

Turned out to need less than I thought. All three of those nested providers import our `TooltipProvider` rather than Radix's, so putting the default on ours reaches them without touching them. One file, no story or test changes.

Skipped the ticket. What a ticket would have covered is the part I left out: a nested provider still shadows the root, so an app-level `delayDuration` will not reach those three. Removing the nested ones makes a root provider a hard requirement, since Radix throws without one, and that is a bigger change than the delay warrants on its own.

PR #629, opened 2026-07-30
feat(tokens): add the missing yellow, lime and red chart colours

jacksonford-07:
Adds the three chart colours the V2 insights frames use but the token set was missing: `special/chart/yellow`, `special/chart/lime` and `special/chart/red`.

## Why

The agency earnings chart plots nine categories. Six had tokens; three were absent, so those series had no legitimate colour to reach for.

Values come straight from the Figma frame's bound variables:

| Token | Value |
|---|---|
| `special/chart/yellow` | `#f5d90a` |
| `special/chart/lime` | `#8fcb3a` |
| `special/chart/red` | `#ff4d4d` |

`lime` and `red` are also what `ChartMetricTrend` reaches for later, which is why they land ahead of it.

## Blast radius

**None.** Purely additive — three new custom properties with no existing consumer, defined identically across the light, `:root` and `.dark` blocks as the rest of the chart ramp is.

## Note on `theme.css`

That file carries an `AUTO-GENERATED — do not edit` header. The change is made in `src/styles/styleTokens.json` and the CSS regenerated with `node src/styles/buildStyles.js`. I re-ran the generator on this branch and it produced no diff, which confirms the committed CSS matches the tokens rather than having been hand-edited.

## Verification

- Full suite — 178 files / 4455 tests passing
- Generator re-run produces no diff

### DIFF
--- a/src/styles/styleTokens.json
+++ b/src/styles/styleTokens.json
@@ -2531,6 +2531,21 @@
                   "exportKey": "variables"
                 }
               }
+            },
+            "yellow": {
+              "type": "color",
+              "value": "#f5d90aff",
+              "blendMode": "normal"
+            },
+            "lime": {
+              "type": "color",
+              "value": "#8fcb3aff",
+              "blendMode": "normal"
+            },
+            "red": {
+              "type": "color",
+              "value": "#ff4d4dff",
+              "blendMode": "normal"
             }
           },
           "default": {
@@ -6172,6 +6187,21 @@
                   "exportKey": "variables"
                 }
               }
+            },
+            "yellow": {
+              "type": "color",
+              "value": "#f5d90aff",
+              "blendMode": "normal"
+            },
+            "lime": {
+              "type": "color",
+              "value": "#8fcb3aff",
+              "blendMode": "normal"
+            },
+            "red": {
+              "type": "color",
+              "value": "#ff4d4dff",
+              "blendMode": "normal"
             }
           },
           "default": {
--- a/src/styles/theme.css
+++ b/src/styles/theme.css
@@ -133,6 +133,9 @@
   --color-special-chart-gray: #c6c6c6ff;
   --color-special-chart-pink: #ff59acff;
   --color-special-chart-purple: #785dffff;
+  --color-special-chart-yellow: #f5d90aff;
+  --color-special-chart-lime: #8fcb3aff;
+  --color-special-chart-red: #ff4d4dff;
   --color-special-default: #9c27b0ff;
   --color-special-background: #fdf1ffff;
   --color-brand-primary-default: var(--primitives-color-green-500);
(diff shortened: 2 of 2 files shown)


src/styles/styleTokens.json
```diff
@@ -2531,6 +2531,21 @@
                   "exportKey": "variables"
                 }
               }
+            },
+            "yellow": {
+              "type": "color",
+              "value": "#f5d90aff",
```
alfonsovo-fv, line 2537: Worth raising with design rather than changing here: this yellow comes out at 1.42:1 against white, the lowest in the chart ramp (`gray` is next at 1.71:1). WCAG 1.4.11 asks for 3:1 on graphics that carry meaning.

As a filled area with a legend that is usually accepted, and the ramp already has this property, so it is not something you are introducing. But as a 1px line on a light background it will be very hard to see. Worth checking whether any series in the earnings chart renders as a thin line rather than a fill, and if so asking for a darker value or a stroke.

alfonsovo-fv (approved):
Approving. The claim worth checking here was the generator one, so I checked it: I re-ran `node src/styles/buildStyles.js` on your branch and `git diff` came back empty, so the committed `theme.css` really does match `styleTokens.json` rather than having been hand-edited past the do-not-edit header. Good instinct to verify that and say so, it is the one thing a reviewer cannot eyeball.

Also confirmed the two details that could have been wrong: the existing ramp order is `teal, sky, magenta, orange, gray, pink, purple`, so not alphabetical and appending matches Figma variable order; and the whole chart ramp is already theme-invariant, so defining the new three identically across all three blocks is the existing pattern rather than an oversight.

One non-blocking note inline about the yellow.

src/styles/styleTokens.json
jacksonford-07: Checked your numbers and they hold: 1.42:1 for this yellow, `gray` next at 1.71:1. Full ramp against white, sorted:

```
1.42  yellow (new)     2.32  sky
1.71  gray             2.47  teal
1.95  lime (new)       2.89  pink
2.11  orange           3.27  red (new)
                       3.29  magenta
                       4.37  purple
```

Answering the question you asked: yes, and it is worse than the ramp position suggests.

`EarningsGraphCard.tsx:201` renders every invoice type as `<Line strokeWidth={2}>`, no fills anywhere in that chart. `CHECKOUT_LINK_SUBSCRIPTION` is the series that takes yellow, via `useEarningsColorMap`.

The part I did not expect: that map already reads `var(--color-special-chart-yellow, #F4C724)`, with the hex as a fallback for the token this PR adds. `#F4C724` is 1.61:1. So this does not introduce a new yellow line, it makes the one eden already draws lighter than it is today. Small regression on a live chart rather than a new risk.

Not changing it here for the reason you gave, plus a mechanical one: `styleTokens.json` is the Design Tokens plugin export of `Color/Special/Chart/Yellow` in the Semantic collection, so editing the value here drifts from the variable and gets overwritten on the next regeneration. It has to change in Figma.

Taking it to Casp with the 1.42 vs 1.61 framing, since "this makes an existing line harder to see" is a more answerable question than "the ramp has a low value in it". One thing I will ask him at the same time: the library has a `Color/Special/Chart/Muted/*` ramp that we do not export at all. If the intent is muted for fills and full for strokes then we may be reaching for the wrong half of the ramp, and I would rather find that out before adding three more.

Resolving this one so the PR is not held on a design question.

PR #630, opened 2026-07-31
fix(CyclingText): read reduced motion during render, split out the motion hook

jacksonford-07:
Follow-up to #621, which stayed a pure move. Two things Alfonso raised there, plus the animation split we discussed.

## Reduced motion is read during render

`usePrefersReducedMotion` now uses `useSyncExternalStore`, as sketched in [his comment](https://github.com/fanvue/fanv-ui/pull/621#discussion_r3689068851). The effect-based version returned `false` on the first render and corrected itself afterwards, so a reduced-motion user could catch one animated frame.

One deviation from the snippet: the `matchMedia` guard stays. `getSnapshot` runs during render, and jsdom does not implement `matchMedia` (`src/test/setup.ts` polyfills `ResizeObserver` but not this), so an unguarded call throws instead of degrading. Verified: `typeof window.matchMedia` is `undefined` in the unit project, which is also why `CyclingText` already had a test named "completes a transition step without relying on matchMedia".

The new test fails against the old implementation and passes against this one. With a faithful stub, where `matches` is a live getter rather than a snapshot, it is the only one of the five that changes verdict. The old hook handled preference changes correctly; the first render was its single defect.

## The slide and cross-fade moved out

Both motion `useMemo`s, `SLIDE_OFFSET_PX` and the easing now live in `useCyclingMotion`, alongside the existing `useCyclingCycle` / `useCyclingTextTrackWidth` / `usePageVisibility`. The hook also owns the reduced-motion read, so `CyclingText` no longer imports `usePrefersReducedMotion` and is left with layout and cycle wiring.

Behaviour is unchanged and the 20 existing `CyclingText` tests pass untouched. It stays inline styles rather than `motion-safe:` utilities because `transitionMs` is a runtime prop.

Adds three reduced-motion tests, a branch that was previously unreachable with no `matchMedia` to stub. Worth noting the cycle is advanced by the fallback timer in `useCyclingCycle`, not by `transitionend`, so it keeps running when no transition fires.

Also drops the commented-out `paddingTop: SLIDE_OFFSET_PX` in `wrapperStyle`, which referenced the constant that moved out.

No public API change; the hook is not exported from `src/index.ts` and `useCyclingMotion` has one internal consumer.

Based on `ds/move-prefers-reduced-motion` so the diff stays clean. GitHub retargets it to `main` once #621 merges.

### DIFF
--- a/src/components/CyclingText/CyclingText.test.tsx
+++ b/src/components/CyclingText/CyclingText.test.tsx
@@ -151,6 +151,54 @@ describe("CyclingText", () => {
     });
   });
 
+  describe("reduced motion", () => {
+    afterEach(() => {
+      Reflect.deleteProperty(window, "matchMedia");
+    });
+
+    const stubReducedMotion = (matches: boolean) => {
+      Object.defineProperty(window, "matchMedia", {
+        configurable: true,
+        writable: true,
+        value: vi.fn((query: string) => ({
+          matches,
+          media: query,
+          onchange: null,
+          addEventListener: () => {},
+          removeEventListener: () => {},
+          addListener: () => {},
+          removeListener: () => {},
+          dispatchEvent: () => false,
+        })),
+      });
+    };
+
+    it("drops the transition when the user prefers reduced motion", () => {
+      stubReducedMotion(true);
+      render(<CyclingText items={ITEMS} intervalMs={500} transitionMs={200} />);
+
+      act(() => vi.advanceTimersByTime(500));
+      expect(getIncomingLabel()).toHaveStyle({ transition: "none" });
+    });
+
+    it("keeps the transition when the user has no preference", () => {
+      stubReducedMotion(false);
+      render(<CyclingText items={ITEMS} intervalMs={500} transitionMs={200} />);
+
+      act(() => vi.advanceTimersByTime(500));
+      expect(getIncomingLabel()?.getAttribute("style")).toContain("200ms");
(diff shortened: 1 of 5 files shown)


alfonsovo-fv (approved):
Approving, and you were right to keep the `matchMedia` guard. My snippet dropped it, and since `getSnapshot` runs during render that would have thrown in the unit project rather than degrading the way the old effect-based version did. Good catch on my suggestion rather than just taking it.

The important detail you got right, which is easy to get wrong: `subscribe` and `getSnapshot` are module-level constants. Inline ones would be new references every render and `useSyncExternalStore` would resubscribe on each, and since `getSnapshot` returns a primitive there is no identity churn either. That is the whole trap with this hook avoided.

Checked the extraction is faithful too. The memo deps go from `[transitionMs, reducedMotion]` to `[durMs]`, which is equivalent because `durMs` is derived from both, and slightly tighter since a `transitionMs` change under reduced motion no longer invalidates a result that cannot change. I also confirmed your note about the cycle: `useCyclingCycle` advances on `setTimeout`, not `transitionend`, so collapsing the duration to zero cannot stall it. 28 tests pass.

One nit, take it or leave it: `getSnapshot` calls `window.matchMedia()` on every render, and each call constructs a new `MediaQueryList`. Caching it at module level behind the guard would avoid that. Small, but this hook runs in every consumer's render path.

PR #631, opened 2026-07-31
fix(Chip): stop a dotted chip resizing when it is selected

jacksonford-07:
Follow-up from [Alfonso's review on #622](https://github.com/fanvue/fanv-ui/pull/622), where he spotted that `dotted` has the same bug the `outlined` variant was fixing.

## Confirmed in a browser first

jsdom cannot do layout, so I measured the real thing in Storybook rather than reasoning about it:

| state | before | after |
| --- | --- | --- |
| dotted unselected | 87.30px | 89.30px |
| dotted selected | 89.30px | 89.30px |

So toggling shifted the row by 2px. Height was 32px throughout, because `h-8` is explicit and `border-box` absorbs the border on a sized axis; only the intrinsic width was affected. Same mechanism as #622.

## The fix

The dashed ring is an SVG, so the unselected chip had no CSS border while the selected one adds 1px. It now reserves a transparent border while unselected, matching the approach in #622.

One extra step that measuring caught: with the border reserved, the SVG was still sized to the padding box, which put the dash 1.5px in from the outer edge while the selected border sits at 0.5px. So they were not on the same ring and toggling would have visibly nudged the outline. The SVG now spans the border box, and the dash lands at 0.5px in both states. The `rx` values were already written for a ring at the outer edge, so they needed no change.

Verified visually in Storybook at both dash patterns (6/6 rounded, 8/8 for the 40px square) and the disabled state. Dash renders cleanly with no clipping from the overhang.

The `asChild` path already draws a real dashed CSS border, so it has no shift and is deliberately untouched. Added a test pinning that.

Based on `ds/chip-outlined` since both touch the same class list; GitHub retargets it to `main` once #622 merges.

### DIFF
--- a/src/components/Chip/Chip.test.tsx
+++ b/src/components/Chip/Chip.test.tsx
@@ -218,6 +218,33 @@ describe("Chip", () => {
       expect(rect).toHaveAttribute("stroke-dasharray", "8 8");
     });
 
+    it("keeps a border in both states", () => {
+      const { rerender } = render(<Chip dotted>New folder</Chip>);
+      expect(screen.getByTestId("chip")).toHaveClass(
+        "border",
+        "border-solid",
+        "border-transparent",
+      );
+
+      rerender(
+        <Chip dotted selected>
+          New folder
+        </Chip>,
+      );
+      expect(screen.getByTestId("chip")).toHaveClass("border", "border-solid");
+    });
+
+    it("does not reserve a border on the asChild path, which draws a real one", () => {
+      render(
+        <Chip dotted asChild>
+          <a href="/x">New folder</a>
+        </Chip>,
+      );
+      const chip = screen.getByRole("link");
+      expect(chip).toHaveClass("border-dashed");
+      expect(chip).not.toHaveClass("border-transparent");
+    });
+
     it("does not render a dashed border by default", () => {
       render(<Chip>Chip</Chip>);
       const chip = screen.getByTestId("chip");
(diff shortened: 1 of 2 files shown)


alfonsovo-fv (approved):
Approving. Measuring this in a browser rather than reasoning about it was the right call, and it caught the thing reasoning would not have: reserving the border fixes the width, but the SVG was still sized to the padding box, so the dash and the solid border it becomes were on different rings. I checked the geometry and it holds. The SVG's containing block is the padding box, so with a 1px border reserved `w-[calc(100%+2px)]` and `-left-px` put it exactly on the border box, and the `rect` at `x=0.5` then lands where a 1px border's centre sits. 42 tests pass.

Two things, neither blocking.

This is branched off `a14611f`, three commits behind #622, so it does not have the `isDotted`/`isSolid`/`isOutlined` refactor and the new rule is written in the old `!isDark && !selected && dotted && !asChild` style. It will conflict when #622 merges, and the risk is resolving it by keeping the old-style condition in a file where everything else now reads `isDotted`. That is the exact inconsistency the refactor was meant to remove, so worth rebasing and rewriting it as `isDotted && !selected && !asChild` rather than fixing it at merge time.

And worth being explicit that this widens the resting state: unselected goes 87.30px to 89.30px, so every existing dotted chip is 2px wider than before, not just stable across toggling. Right fix, but there are 7 dotted usages in the repo and Chromatic will diff all of them. Fine, just call it out in the description so the diffs are expected rather than surprising.

PR #633, opened 2026-07-31
fix(tooltip): default TooltipProvider delayDuration to 200ms

jacksonford-07:
Follow-up to the `delayDuration` thread on #628.

Radix's `Tooltip.Provider` defaults to 700ms. On the short hint labels this library uses tooltips for (`TooltipLabel`, the `ChartCard` info icon, `SwitchField`), that reads as the tooltip not working rather than as a deliberate delay. This sets our `TooltipProvider` to 200ms by default, still overridable per provider.

## Scope

Deliberately one file. `TooltipLabel`, `ChartCard` and `SwitchField` each mount their own `<TooltipProvider>` internally, and they import ours rather than Radix's, so they pick up the new default for free. No component, story or test needed a change.

## What this does not do

A nested provider still shadows the root one, so an app that sets its own `delayDuration` at the root will not reach those three components. Fixing that means removing the nested providers, which makes a root `TooltipProvider` a hard requirement (Radix throws `Tooltip must be used within TooltipProvider` without one) and touches every test and story that renders them. That is a separate change and not needed to fix the delay.

## Verification

Two tests, both driving the real Radix timer: closed at 199ms, open at 200ms. The 200ms case fails against the previous bare re-export and passes with the change. Full suite 4502/4502, `tsc --noEmit` clean, biome clean.

### DIFF
--- a/src/components/Tooltip/Tooltip.test.tsx
+++ b/src/components/Tooltip/Tooltip.test.tsx
@@ -1,7 +1,7 @@
-import { render, screen } from "@testing-library/react";
+import { act, fireEvent, render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import * as React from "react";
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { axe } from "vitest-axe";
 import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "./Tooltip";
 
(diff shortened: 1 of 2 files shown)


alfonsovo-fv (approved):
Approving. Small, and scoped exactly right. 23 tests pass.

The two timer tests are the good part: driving Radix's real timer and asserting closed at 199ms, open at 200ms, means they genuinely fail against the old bare re-export. A test that just asserted the prop was passed would not have.

Also the right call to document the nested-provider caveat in the doc comment rather than quietly fixing it here. Removing those three internal providers makes a root `TooltipProvider` a hard requirement and touches every story and test that renders them, which is not something to smuggle into a delay change. Worth a ticket so it does not get forgotten, since the surprise for an app author is real: they set `delayDuration` at the root and three components ignore it.

One thing worth a line in the description: this changes the delay for all 62 existing tooltip consumers, and as a `fix` it ships on a patch. I think `fix` is defensible, 700ms genuinely reads as broken rather than deliberate, but it is the same shape as the default change on #627, so give it the same treatment and mention it in the release notes rather than letting people notice.

PR #634, opened 2026-07-31
feat(TrendPill): add the filled trend pill from the V2 design

jacksonford-07:
First of six extracted from the `insights-components` batch. Each one is independent and can be reviewed in any order.

`TrendPill` is Figma's `V2 Trend Pill`: a filled pill stating a change, green with a plus for positive, red with a minus for negative. Agency insights uses it inside list rows.

It is deliberately not the same thing as `ChartCard`'s `trendChip`, which is Figma's `V2 Trend` (unfilled arrow beside a headline). The doc comment says so, because the two are easy to reach for interchangeably and they are not interchangeable: this one has to hold its own against surrounding content rather than trail a number.

Built on the existing `Pill` and the existing `PlusIcon` / `MinusIcon`, so there is no new primitive here.

Also adds the `App.tsx` demo per `AGENTS.md`. The batch branch had skipped that for the new components.

9 tests including axe, `tsc --noEmit` clean, biome clean.

### DIFF
--- a/src/components/TrendPill/TrendPill.tsx
+++ b/src/components/TrendPill/TrendPill.tsx
@@ -12,6 +12,13 @@ export interface TrendPillProps extends Omit<React.HTMLAttributes<HTMLSpanElemen
   label: React.ReactNode;
   /** Drives the colour and the leading glyph. */
   trend: TrendDirection;
+  /**
+   * Announced before {@link label} when `trend` is `"positive"`. Overridable
+   * because the component cannot translate it. @default "Increase"
+   */
+  increaseLabel?: string;
+  /** As {@link increaseLabel}, for `"negative"`. @default "Decrease" */
+  decreaseLabel?: string;
 }
 
 /**
@@ -29,7 +36,7 @@ export interface TrendPillProps extends Omit<React.HTMLAttributes<HTMLSpanElemen
  * ```
  */
 export const TrendPill = React.forwardRef<HTMLSpanElement, TrendPillProps>(
-  ({ label, trend, ...props }, ref) => {
+  ({ label, trend, increaseLabel = "Increase", decreaseLabel = "Decrease", ...props }, ref) => {
     const isPositive = trend === "positive";
 
     return (
@@ -39,6 +46,13 @@ export const TrendPill = React.forwardRef<HTMLSpanElement, TrendPillProps>(
         leftIcon={isPositive ? <PlusIcon /> : <MinusIcon />}
         {...props}
       >
+        {/*
+         * `Pill` renders `leftIcon` inside `aria-hidden`, so without this the only
+         * channel carrying the direction is the green/red fill and the label itself
+         * says nothing about which way it went — WCAG 1.4.1, and the direction is
+         * the point of a trend.
+         */}
+        <span className="sr-only">{isPositive ? increaseLabel : decreaseLabel}</span>
         {label}
       </Pill>
     );
--- a/src/components/TrendPill/TrendPill.tsx
+++ b/src/components/TrendPill/TrendPill.tsx
@@ -1,0 +1,48 @@
+import * as React from "react";
+import { MinusIcon } from "../Icons/MinusIcon";
+import { PlusIcon } from "../Icons/PlusIcon";
+import { Pill } from "../Pill/Pill";
+
+/** Whether a change is an improvement or a decline. */
+export type TrendDirection = "positive" | "negative";
+
+/** Props for {@link TrendPill}. */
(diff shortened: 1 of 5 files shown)


jacksonford-07:
https://www.figma.com/design/S8zFdcOjt4qN4PrwntuCdt/Fanvue-Library?node-id=20558-588&m=dev

jacksonford-07:
Checked against the link, and I think it points at the wrong component — worth resolving before Alfonso looks, because it is the exact confusion this PR's doc comment warns about.

`20558:588` renders as `↗ X% vs prev` green and `↘ X% vs prev` red: unfilled, diagonal arrows, 14px Regular. Its variables are `Color/Success/Content` (#0b8e29), `Color/Error/Content` (#f01932) and `Border/Primary`, with no fill token at all.

That is already built. `ChartCard`'s `trendChip` on `main` is a byte-for-byte match:

- `TREND_CLASSES` is `text-success-content` / `text-error-content` — the same two tokens
- `ArrowUpRightIcon` with `rotate-90` for negative — the ↗ / ↘ pair
- `typography-body-small-14px-regular` — the same 14px Regular
- no background

So `20558:588` is `V2 Trend`, not `V2 Trend Pill`.

`V2 Trend Pill` does exist separately — component set `40a87bc62dd0a6dea0f9c3e5854b99e91194de9b` in the Fanvue Library, last updated 16 Jul — and that is what this PR implements: filled green/red with plus and minus glyphs. So I do not think anything here is wrong, but I could not verify it visually because search returns the component key without a node id, and this file only exposes one page to the MCP so I cannot navigate to it.

Could you drop the `V2 Trend Pill` node link instead? Then I can confirm the fill, the glyphs and the type scale properly rather than asserting them.

If the intent was the opposite — that agency insights wants the unfilled treatment here and `TrendPill` is redundant — say so and I will close this and use `trendChip`, which would be the better outcome since it is already shipped.

src/components/TrendPill/TrendPill.tsx
```diff
@@ -0,0 +1,48 @@
...
+ */
+export const TrendPill = React.forwardRef<HTMLSpanElement, TrendPillProps>(
+  ({ label, trend, ...props }, ref) => {
+    const isPositive = trend === "positive";
+
+    return (
+      <Pill
+        ref={ref}
+        variant={isPositive ? "green" : "red"}
+        leftIcon={isPositive ? <PlusIcon /> : <MinusIcon />}
```
alfonsovo-fv, line 46: The direction is only available visually. `Pill` wraps `leftIcon` in `aria-hidden="true"` (`Pill.tsx` line 94), so the glyph is not exposed, and the only other channel is the green/red fill. A screen reader gets "$240 vs Jun" with nothing saying whether that is up or down, and the label itself carries no sign.

That is WCAG 1.4.1 as well as plain information loss: for a trend, the direction is the point. Needs a text channel, ideally overridable for i18n the way you did `loadingLabel` on #627:

```tsx
<Pill ...>
  <span className="sr-only">{trend === "positive" ? increaseLabel : decreaseLabel}</span>
  {label}
</Pill>
```

Worth noting the axe test passes and always will here, for the same reason it did on the skeleton: nothing in the markup is invalid, the information is just absent.

alfonsovo-fv (commented):
One real issue, otherwise this is a clean wrapper.

src/components/TrendPill/TrendPill.tsx
jacksonford-07: Fixed. `increaseLabel` / `decreaseLabel` props, defaulting to "Increase" / "Decrease", rendered in an `sr-only` span before the label, following `loadingLabel` on #627.

Confirmed the cause rather than taking it: `Pill.tsx:94` wraps `leftIcon` in `aria-hidden="true"`, so the glyph was never exposed and the fill was the only channel left.

Your closing point is the one I acted on hardest. Two tests, neither of them axe:

```
✓ announces the direction, which colour and glyph alone do not carry
✓ lets the host translate the direction
```

With a comment saying why axe cannot cover it, because that is now the second time in this batch a passing axe run has meant "the markup is valid" rather than "the information is there".

Went with `sr-only` over `aria-label` on the pill because the label is already text: an `aria-label` would replace "$240 vs Jun" rather than prefix it, so the figure would have to be duplicated into the label string by every caller.

alfonsovo-fv (approved):
Approving. The `sr-only` span with overridable `increaseLabel`/`decreaseLabel` is exactly right, and the two tests assert the text channel rather than leaning on axe. Noting in the test file that axe passes either way is a good habit to keep.

PR #635, opened 2026-07-31
feat(AnimatedNumber): add the roll and count value animations

jacksonford-07:
Second of six extracted from the `insights-components` batch. Independent of the others.

Animates a number when its value changes, for the insights headline figures. Two variants because one does not fit both cases:

- `roll` is a per-digit odometer, each column sliding to its new digit with the box width easing when the digit count changes. Mechanical, so it suits a single headline figure.
- `count` interpolates the value and re-formats every frame, so the whole number counts up. Calmer when several figures on a page change at once.

`format` receives interpolated values in the `count` variant, so it has to handle non-integers. That is documented on the prop rather than left to be discovered.

Honours `usePrefersReducedMotion`, which is why this sits after #621 and #630 rather than before them.

Adds the `App.tsx` demo per `AGENTS.md`, showing both variants side by side so the difference is visible rather than described.

11 tests including axe, `tsc --noEmit` clean, biome clean.

### DIFF
--- a/src/components/AnimatedNumber/AnimatedNumber.tsx
+++ b/src/components/AnimatedNumber/AnimatedNumber.tsx
@@ -1,0 +1,238 @@
+import * as React from "react";
+import { cn } from "@/utils/cn";
+import { usePrefersReducedMotion } from "@/utils/usePrefersReducedMotion";
+
+/** How a value change is animated. */
+export type AnimatedNumberVariant =
+  /**
+   * Per-digit odometer: each digit column slides to its new value and the box
+   * width eases when the digit count changes. Reads as mechanical, so it suits
+   * a single headline figure rather than a grid of them.
+   */
+  | "roll"
+  /**
+   * Interpolates the value and re-formats it each frame, so the whole number
+   * counts from old to new. Calmer than `roll` when several figures on a page
+   * change at once.
+   */
+  | "count";
+
+/** Props for {@link AnimatedNumber}. */
+export interface AnimatedNumberProps
+  extends Omit<React.HTMLAttributes<HTMLSpanElement>, "children"> {
+  /** The value to display. Animates whenever this changes. */
+  value: number;
+  /**
+   * Turns the value into the displayed string. Receives interpolated values in
+   * the `count` variant, so it must handle non-integers.
+   * @default String(Math.round(value))
+   */
+  format?: (value: number) => string;
+  /** Which animation to run. @default "count" */
+  variant?: AnimatedNumberVariant;
+  /** Animation duration in milliseconds. @default 500 */
+  durationMs?: number;
+}
+
+const DEFAULT_DURATION_MS = 500;
+
+const defaultFormat = (value: number) => String(Math.round(value));
+
+/** Decelerating ease, so the value settles rather than stopping abruptly. */
+const easeOutCubic = (t: number) => 1 - (1 - t) ** 3;
+
+const DIGITS = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"];
+
+const isDigit = (char: string) => char >= "0" && char <= "9";
+
+/**
+ * Animates a number when its value changes, either by rolling each digit like
+ * an odometer or by counting through the intermediate values.
+ *
+ * Both variants collapse to an instant swap under
+ * `prefers-reduced-motion: reduce`.
+ *
+ * The rolling variant stacks all ten digits in each column, so the visual is
+ * hidden from assistive technology and the formatted value is exposed once as
+ * text instead.
+ *
+ * @example
+ * ```tsx
+ * // Headline figure
+ * <AnimatedNumber value={totalGross} format={formatPrice} variant="roll" />
+ *
+ * // Supporting figure
+ * <AnimatedNumber value={activeFans} variant="count" />
+ * ```
+ */
+export const AnimatedNumber = React.forwardRef<HTMLSpanElement, AnimatedNumberProps>(
+  (
+    {
+      value,
+      format = defaultFormat,
+      variant = "count",
+      durationMs = DEFAULT_DURATION_MS,
+      className,
+      ...props
+    },
+    ref,
+  ) => {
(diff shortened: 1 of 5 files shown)


src/components/AnimatedNumber/AnimatedNumber.tsx
```diff
@@ -0,0 +1,238 @@
...
+          frame = requestAnimationFrame(step);
+        } else {
+          fromRef.current = value;
+        }
+      };
+
+      frame = requestAnimationFrame(step);
+      return () => {
+        cancelAnimationFrame(frame);
+        fromRef.current = value;
```
alfonsovo-fv: This makes an interrupted tween jump. Setting `fromRef` to `value` in the cleanup stores the *target* being abandoned, not where the display actually got to, so the next tween starts from a number that was never on screen.

Drove rAF manually to confirm:

```
mount:                       0
mid-tween 0->100 (~100ms):  49
retarget to 200:            49
first frame after retarget: 100   <- snaps to the abandoned target
+100ms:                     149
```

So a value change while animating shows a visible jump to the old target before counting on. For insights headline figures that is a refetch or a filter change landing mid-animation, which is not exotic.

The fix is to resume from what is displayed. Track it in a ref alongside the state and read that as `from`:

```ts
const displayRef = React.useRef(value);
// in step(): displayRef.current = next; setDisplayValue(next);
// on the next run: const from = displayRef.current;
```

and drop the `fromRef.current = value` from the cleanup entirely.

Worth saying the comment above it describes a different problem than the code solves. Guarding against a re-render restarting the tween is not needed, since the effect only re-runs when `value` or `durationMs` change. What the cleanup actually does is decide where an interrupted tween resumes, and it picks the wrong point.

src/components/AnimatedNumber/AnimatedNumber.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+    React.useLayoutEffect(() => {
+      if (!display) {
+        return;
+      }
+      const measured = contentRef.current?.scrollWidth;
+      if (measured) {
+        setBoxWidth(measured);
+      }
+    }, [display]);
```
alfonsovo-fv, line 202: The box width is measured once per `display` and never re-measured, and the outer span is `overflow-hidden`. So if the web font loads after this runs, the locked pixel width is too small for the wider glyphs and the number is clipped until the value next changes, which for a headline figure may be never.

This repo already treats that as worth handling in two places, both measuring text: `useCyclingTextTrackWidth` attaches a `ResizeObserver` and its doc comment says explicitly "when fonts load or the parent resizes", and `TabsList` waits on `document.fonts.ready`. A `ResizeObserver` on the content span would cover both the font case and a container change.

Separately, the `hasMeasured` guard does not do what its comment says. `boxWidth` is set by the layout effect, so on the render where the width first becomes a pixel value `hasMeasured` is already `true` and the full duration applies. The guard only zeroes the duration on the render where the width is still `auto` and there is nothing to transition from. It happens not to matter, because `auto` to a length is not interpolatable so nothing animates either way, but the comment claims a protection that is not there.

alfonsovo-fv (changes requested):
Two real issues in the animation paths, one of which I reproduced. The rest of this is good, and not animating on mount in the `count` variant is the right default.

src/components/AnimatedNumber/AnimatedNumber.tsx
jacksonford-07: Fixed exactly as you set out: `displayRef` tracks what is on screen, `step` writes to it every frame, the next run reads it as `from`, and the `fromRef.current = value` is gone from the cleanup.

Your reading of the comment was right too, and it was the tell. It claimed to stop a re-render restarting the tween, which the dependency array already does. Rewritten to describe what the ref is actually for.

Test asserts your repro:

```
retarget mid-flight, advance one frame
→ must be >= the mid value and still < 100
```

It fails against the old code and passes with the fix.

One thing your case surfaced that I did not expect: the test harness was broken and had been passing only because nothing exercised the cleanup with a frame pending. `beforeEach` hand-stubbed `requestAnimationFrame` over `setTimeout` while `cancelAnimationFrame` cleared with `clearTimeout`, and the first test to interrupt a tween got:

```
Error: Cannot clear timer: timer created with setTimeout() but cleared with cancelAnimationFrame()
```

Vitest fakes `requestAnimationFrame` natively, so the stubs are gone and `vi.useFakeTimers()` alone drives it. All 12 pass, full suite 4538.

src/components/AnimatedNumber/AnimatedNumber.tsx
jacksonford-07: Both parts fixed.

The measurement is now a `ResizeObserver` on the content span, so a font landing late re-locks the width instead of clipping until the next value change. Followed `useCyclingTextTrackWidth` since you pointed at it, including the `typeof ResizeObserver === "undefined"` guard.

The `hasMeasured` comment is rewritten to say what it does rather than what it protects against. You are right that it fires on the `auto` render, not the first pixel render, and that it happens not to matter because `auto` to a length is not interpolatable. It now says that outright and calls itself belt-and-braces, rather than claiming a protection it does not provide.

Both of these were the same failure on my side: a comment asserting a mechanism I had not checked. That is now twice in this file.

alfonsovo-fv (approved):
Both fixed, approving. I re-ran the manual rAF probe:

```
mid-tween 0->100:           49
first frame after retarget: 49   (was 100)
+100ms:                     123
settled:                    200
```

So an interrupted tween now resumes from what is on screen instead of snapping to the abandoned target. Dropping the cleanup assignment entirely rather than patching what it stored is the cleaner shape.

The `ResizeObserver` covers the font case and a container change, and the observer settles rather than looping, since a same-value `setBoxWidth` bails out of the re-render. You also corrected the `hasMeasured` comment to say it is belt-and-braces rather than load-bearing, which is what it is.

One trivial thing: the description still says 11 tests, the suite runs 9. Worth keeping that line accurate for the same reason as the `size-limit` one on #627.

PR #636, opened 2026-07-31
feat(AiButton): add the shimmering AI action button

jacksonford-07:
Third of six extracted from the `insights-components` batch. Independent of the others.

The AI action button from the insights designs: a per-letter shimmer at rest, a glyph flicker, and a bottom sheen that tracks the cursor. `active` swaps to `activeLabel` and keeps the shimmer running while work is in flight.

Two things worth a reviewer's eye:

**The letter run is hidden from assistive tech.** It renders one `<span>` per character to give each its own animation delay, which a screen reader would otherwise spell out. The run carries `aria-hidden` and the accessible name comes from `label` (or `aria-label`), so it says "Analyse" rather than "A, n, a, l, y, s, e".

**The 24 size steps its label down to the 12px scale**, unlike `Button`, which uses 14px at that height. Not a mismatch: 16px is the smallest authored icon geometry and cannot step down with the label, so at 14px the glyph and text crowd a 24px box. The 24 button renders the 16px glyph optically at 14 for the same reason. Both are commented at the constant.

The cursor-tracking sheen writes its x offset straight to the DOM node rather than through state, deliberately, so a pointer move does not re-render the button.

Honours `motion-reduce`.

**Three biome warnings come across with the code** and I have left them rather than widening the diff: two `useSortedClasses` inside arbitrary-value arrays where reordering risks changing specificity, and one `noArrayIndexKey` on the character run, where the index genuinely is the identity because characters repeat. Happy to add a `biome-ignore` with that rationale if you would rather it were explicit. Neither fails lint (`biome check` exits 0 on warnings) and `App.tsx` already ships two of the same kind.

24 tests including axe, `tsc --noEmit` clean.

### DIFF
--- a/src/components/AiButton/AiButton.tsx
+++ b/src/components/AiButton/AiButton.tsx
@@ -67,20 +67,26 @@ export interface AiButtonProps extends Omit<React.ButtonHTMLAttributes<HTMLButto
  * The run is hidden from assistive tech — the button's accessible name comes from
  * its `aria-label`, so a screen reader says "Analyse" rather than spelling it out.
  */
-const ShimmerLabel: React.FC<{ text: string }> = ({ text }) => (
+const ShimmerLabel: React.FC<{ text: string; disabled?: boolean }> = ({ text, disabled }) => (
   <span aria-hidden="true">
     {Array.from(text).map((character, index) => (
       <span
         key={`${character}-${index}`}
         className={cn(
           "inline-block whitespace-pre text-content-secondary",
-          "[animation:fv-ai-letter_2s_ease-in-out_infinite]",
-          "group-hover/ai:text-content-primary group-hover/ai:[animation:none]",
-          // On focus each letter blows up and settles, then the idle shimmer
-          // resumes underneath it — the two animations are sequenced by delay.
-          "group-focus-visible/ai:[animation:fv-ai-letter-focus_1s_ease-in-out,fv-ai-letter_1.2s_ease-in-out_infinite_1s]",
-          "group-active/ai:text-content-primary group-active/ai:[animation:none]",
-          "group-active/ai:[text-shadow:0_0_4px_var(--color-content-primary)]",
+          // Gated in JS rather than with a `group-disabled/ai:` override, because
+          // that would rely on the generated rule ordering to beat `group-hover`.
+          // A faded control that still shimmers and still lights up green under the
+          // cursor reads as busy rather than unavailable.
+          !disabled && [
+            "[animation:fv-ai-letter_2s_ease-in-out_infinite]",
+            "group-hover/ai:text-content-primary group-hover/ai:[animation:none]",
+            // On focus each letter blows up and settles, then the idle shimmer
+            // resumes underneath it — the two animations are sequenced by delay.
+            "group-focus-visible/ai:[animation:fv-ai-letter-focus_1s_ease-in-out,fv-ai-letter_1.2s_ease-in-out_infinite_1s]",
+            "group-active/ai:text-content-primary group-active/ai:[animation:none]",
+            "group-active/ai:[text-shadow:0_0_4px_var(--color-content-primary)]",
+          ],
           "motion-reduce:[animation:none]",
         )}
         style={{ animationDelay: `${index * LETTER_STAGGER}s` }}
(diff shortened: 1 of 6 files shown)


jacksonford-07:
image
 not in figma but will wanted it, will either allign it to figma or get someone to create a new variant for it

jacksonford-07:
Fixed in `df09fbc`, and found a second thing while in there.

**The icon colour.** It was `text-content-always-white` on a `bg-background-primary` button. `--color-background-primary` is `gray-white` in light mode, so the glyph was white on white. Now `text-content-primary`, which resolves to `gray-black` in light and `gray-white` in dark (`theme.css:46` and `:657`).

Same bug one line down, so I took it in the same pass: the resting glow was `drop-shadow(... content-always-white 60% ...)` and the press glow was `text-shadow: 0 0 4px var(--color-content-always-white)`. A white glow behind a black glyph is invisible for the same reason. Both now follow `content-primary`.

**The bigger one: none of the animation actually ran.** `4264801`.

The component references `fv-ai-button`, `fv-ai-letter`, `fv-ai-flicker`, `fv-ai-letter-focus` and `--fv-ai-x`. All five live in `base.css` on the batch branch and I had only extracted `src/components/AiButton/`, so this PR was shipping a button whose shimmer, glyph flicker, focus bloom, lit rim and cursor-tracking sheen were all no-ops. The doc comment even says the rim and sheen come from that utility.

The tests did not catch it and could not: jsdom does not evaluate CSS, so 24 passing tests say nothing about whether a referenced keyframe exists. Worth noting for the other five PRs in this set, though I have checked and none of them reference a utility that main lacks.

166 lines of `base.css` added. `pnpm build` now passes with the CSS compiling, which is the check that would have caught this.

**One thing I have deliberately left.** The rim and sheen inside `fv-ai-button` still use `content-always-white` for their inset box-shadows, and the mask uses it too. The mask is fine, only its alpha matters. The rim is a judgment call I did not want to make silently: a white lit edge reads on a dark button and largely disappears on a light one, so the whole hover treatment may want rethinking for light mode rather than a token swap. That is a design question, not a token bug, so flagging rather than changing.

src/components/AiButton/AiButton.tsx
```diff
@@ -0,0 +1,206 @@
...
+ * its `aria-label`, so a screen reader says "Analyse" rather than spelling it out.
+ */
+const ShimmerLabel: React.FC<{ text: string }> = ({ text }) => (
+  <span aria-hidden="true">
+    {Array.from(text).map((character, index) => (
+      <span
+        key={`${character}-${index}`}
+        className={cn(
+          "inline-block whitespace-pre text-content-secondary",
+          "[animation:fv-ai-letter_2s_ease-in-out_infinite]",
```
alfonsovo-fv: A disabled button keeps shimmering. There is no `group-disabled/ai:` guard on either animation, so with `disabled` set the letters still run `fv-ai-letter` and the glyph still runs `fv-ai-flicker` (line 171) under `disabled:opacity-60`. A faded control that is still animating reads as busy rather than unavailable, and it is the one state where the motion is actively misleading.

Since the button is the group, it is one class in each place:

```
"group-disabled/ai:[animation:none]"
```

Worth checking the hover treatment too: `disabled:hover:border-border-primary` holds the outline, but the letters' `group-hover/ai:text-content-primary` and the glyph's `group-hover/ai:` drop-shadow still fire on a disabled button, so it lights up green on hover while unavailable.

alfonsovo-fv (commented):
The a11y thinking here is good: `aria-hidden` on the per-letter run with the name on the button, `aria-busy` while active, and `motion-reduce` on every keyframe animation. I checked the `fv-ai-letter` keyframes and the un-animated base state is the resting colour, so reduced motion genuinely lands settled rather than blank.

One real issue.

src/components/AiButton/AiButton.tsx
jacksonford-07: Fixed, and I took the hover half you flagged at the end as well.

Gated in JS rather than with `group-disabled/ai:[animation:none]`. The animation override would have worked, but the colour and filter ones would have been racing `group-hover/ai:` on generated rule order, and `disabled` is already destructured two lines up. So when disabled the letters emit no animation, no `group-hover:text-content-primary`, no focus and active rules; the glyph emits no flicker and no green drop-shadow. `motion-reduce:[animation:none]` stays outside the guard since it applies either way.

`ShimmerLabel` takes a `disabled` prop now, passed at both call sites so the active label behaves the same.

Two tests, the pair rather than one:

```
✓ stops shimmering and stops lighting up on hover once disabled
✓ still shimmers when enabled
```

The second is there because the first would pass against a component that never animates at all.

`disabled:hover:border-border-primary` was already holding the outline, so that part was fine — it was only the letters and glyph reaching past it.

alfonsovo-fv (approved):
Approving, and gating in JS was the better call than the `group-disabled/ai:` override I suggested. Both are same-specificity arbitrary variants, so which one wins would have come down to generated rule order, which Tailwind does not guarantee. Conditioning the class list removes the question instead of betting on it.

Good that you took the hover half too, not just the animation. A disabled button lighting up green under the cursor was the same bug wearing a different hat.

PR #637, opened 2026-07-31
feat(Chart): add ChartMetricTrend for the metric tile area chart

jacksonford-07:
Fourth of six extracted from the `insights-components` batch. Independent of the others.

The small area chart that fills the `V2 Area Chart Item` slot of the Figma insight card. Exported from `@fanvue/ui/charts`, not the root, matching `ChartSkeleton` and `ChartPieLegend`. Story only, no `App.tsx` demo, for the same reason.

Two decisions worth checking rather than skimming:

**`series` is optional and absence draws a flat line.** The endpoints behind these cards often return a single scalar with no history. Rather than collapse the slot and lose the height and shape the design gives the card, it draws a level line at the current value. Level rather than shaped on purpose, so it asserts no trend the data cannot support. That is the kind of thing worth disagreeing with if you do, so it is spelled out in the doc comment.

**`green` maps to `teal`.** The design puts red on Retention Rate and green on Average Fan Lifetime, but there is no `special-chart-green` token. `teal` (`#28ba8e`) is the emerald in the Figma render; `lime` (`#8fcb3a`) is yellow-green and would be wrong. Noted at the type.

This is the first consumer of `special-chart-red` from #629. Relevant to your contrast note there: red is 3.27:1 against white, so it clears WCAG 1.4.11's 3:1 even as the stroke it is used as here. The yellow at 1.42:1 is not used by this component.

10 tests including axe, `tsc --noEmit` clean, biome clean with no fixes needed.

### DIFF
--- a/src/components/Chart/ChartMetricTrend.tsx
+++ b/src/components/Chart/ChartMetricTrend.tsx
@@ -1,0 +1,88 @@
+import * as React from "react";
+import { Area, AreaChart, YAxis } from "recharts";
+import { cn } from "../../utils/cn";
+import { ChartContainer } from "./ChartContainer";
+import type { ChartConfig } from "./types";
+
+/**
+ * Each card carries its own gradient in the design: red on Retention Rate,
+ * green on Average Fan Lifetime. The DS has no `special-chart-green`, so the
+ * green maps to `teal` (#28ba8e) — the emerald in the Figma render, as opposed
+ * to `lime` (#8fcb3a), which is yellow-green.
+ */
+export type ChartMetricTrendColor = "red" | "teal";
+
+const CHART_COLORS: Record<ChartMetricTrendColor, string> = {
+  red: "var(--color-special-chart-red)",
+  teal: "var(--color-special-chart-teal)",
+};
+
+/** Props for {@link ChartMetricTrend}. */
+export interface ChartMetricTrendProps extends React.HTMLAttributes<HTMLDivElement> {
+  /** Current value. Also the flat-line height when `series` is absent. */
+  value: number;
+  /** Gradient and stroke colour. */
+  color: ChartMetricTrendColor;
+  /** History to plot. Needs more than one point to draw a shape. */
+  series?: number[];
+}
+
+/**
+ * The small area chart inside a metric tile, per the `V2 Area Chart Item` slot of
+ * the Figma insight card.
+ *
+ * `series` is optional because the endpoints behind these cards often return a
+ * single scalar with no history. With nothing to plot the chart draws a flat line
+ * at the current value rather than collapsing the slot, so the card keeps the
+ * height and shape the design gives it. The line is deliberately level rather
+ * than shaped, so it asserts no history that the data cannot support.
+ *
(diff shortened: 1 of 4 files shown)


jacksonford-07:
@alfonsovo-fv reason i added these is currently with retention metrics, there are no values for the line charts, so a fix was to add these until enough historical data is made up, then we can import and use the corrected ones when the prop has enough data to pass into it. Needs a second pair of eyes tho

src/components/Chart/ChartMetricTrend.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * <ChartMetricTrend value={42} color="teal" />
+ * ```
+ */
+export const ChartMetricTrend = React.forwardRef<HTMLDivElement, ChartMetricTrendProps>(
+  ({ value, color, series, className, ...props }, ref) => {
+    const points = series && series.length > 1 ? series : [value, value];
+    const chartData = points.map((point, index) => ({ index, value: point }));
+    const chartConfig: ChartConfig = { value: { color: CHART_COLORS[color] } };
+    // Scoped per colour so two cards on the same page don't share one gradient def.
+    const gradientId = `metric-trend-gradient-${color}`;
```
alfonsovo-fv: The comment says this is scoped so two cards do not share a gradient def, but scoping by colour is exactly what makes two same-colour cards share one: both emit `id="metric-trend-gradient-teal"` and every `url(#...)` in the document resolves to whichever comes first.

It is not visible today, because same-colour gradients are identical and the design pairs one red with one green. But duplicate ids in a document are the kind of thing that bites later, and `React.useId()` makes it per-instance for free:

```ts
const gradientId = `metric-trend-gradient-${React.useId()}`;
```

Then the comment is true as written.

alfonsovo-fv (approved):
Approving. The flat-line fallback with the padded `YAxis` domain is a nice touch, keeping a single scalar inside the plot area rather than reading as zero, and `isAnimationActive={false}` sidesteps the reduced-motion question entirely.

One thing worth tidying, not blocking.

src/components/Chart/ChartMetricTrend.tsx
jacksonford-07: Fixed with `React.useId()`, and you are right that the comment was describing the opposite of what the code did.

One addition to your snippet: `useId` emits colons (`:r0:`), which are legal in an `id` and fine inside `url(#...)`, but break `querySelector`. Stripped them so the id stays usable either way.

```ts
const gradientId = `metric-trend-gradient-${React.useId().replace(/:/g, "")}`;
```

Comment rewritten to state the collision it now actually prevents.

Could not cover it in the unit test: recharts needs a measured container and jsdom reports zero size, so there is no `<defs>` in the DOM to assert on. That is already called out in the file's scope note — but the note claimed the story covered "per-colour gradient scoping", and no story rendered two same-colour cards, so the coverage it pointed at did not exist. Added `TwoSameColourCards` (two teal, opposite shapes), which is the case that would have broken, and corrected the note to say per-instance.

PR #638, opened 2026-07-31
feat(UserItem): add trailing, primary and secondary slots plus UserItemTrailing

jacksonford-07:
Fifth of six extracted from the `insights-components` batch. The only one of the six that touches an existing component, so it is the one with real blast radius: `UserItem` has 15 consumers across `eden` and `fadmin`.

All three new props are optional and nothing renders differently without them, which is why the full suite passes unchanged (4545 tests). Worth confirming that claim rather than taking it, since it is the whole safety argument for this PR.

**What it adds**

- `trailing` pins content to the end of the row. It keeps its intrinsic width while the name block absorbs the remaining space, so figures stay aligned down a list rather than drifting with name length.
- `UserItemTrailing` is the thing to put in that slot: a value with an optional `meta` line, right-aligned.
- `primary` replaces the display-name line for rows that read as a sentence about the user rather than their name, such as an activity feed entry. Badges deliberately do not apply to custom content.
- `secondary` adds a supporting line under the name, mirroring `UserItemTrailing`'s `meta`. Both sit on the same 18px advance, so a row using `secondary` on the left and `meta` on the right reads as two aligned tiers.

`secondary` exists specifically to avoid letting a long primary wrap, which mixes line heights and drifts the rows out of step. That is documented on the prop.

Serves top spenders, most active fans and the recent-activity feed in agency insights.

57 tests in `UserItem` including axe, full suite 4545/4545, `tsc --noEmit` clean.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -265,6 +265,8 @@ import {
   UserCircleIcon,
   UserDisplayName,
   UserIcon,
+  UserItem,
+  UserItemTrailing,
   UsersIcon,
   VideoIcon,
   VipBadgeIcon,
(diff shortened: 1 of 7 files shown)


jacksonford-07:
this is a more temporary cleanup, we can amend the table if needed when we refresh creator insights at a later date

jacksonford-07:
temp fix, the current ones in eden use inconsistent padding, and icon sizes making creator earnings dashboard look very off, lmk if you think it's better to just import directly into eden with it being temp

alfonsovo-fv (approved):
Approving, nothing to flag. All three additions are opt-in and I checked existing consumers are untouched: `primary ?? <UserDisplayName>` keeps the current render when the prop is absent, and `secondary` and `trailing` are plain `&&` guards.

`UserItemTrailing` earning its own component rather than being composed at each call site is the right instinct, since the whole point is that figures line up down a list, and that only holds if the typography is decided once.

PR #639, opened 2026-07-31
feat(DropdownMenu): add DropdownMenuCheckboxItem and align the menu with V2

jacksonford-07:
Last of six extracted from the `insights-components` batch. It is the one bundle in the set, because all four changes land in the same file and serve the same design. Reviewing them as four separate concerns:

**1. `DropdownMenuCheckboxItem`** — a multi-select row: square indicator that fills when checked, optional leading avatar, optional helper line. This is the new export. Used by the agency insights creator filter.

**2. A search header can take focus on open.** `DropdownMenuHeader` gains `searchProps.autoFocus`. Focus is applied after mount rather than through the `autoFocus` attribute, because Radix moves focus to the content on open and the attribute loses that race.

**3. No focus ring left on the trigger after a click-away.** Radix returns focus to the trigger on close, which is right for Escape and wrong for a pointer dismiss: clicking outside left a visible ring on a button the user had moved away from. Now `onPointerDownOutside` records that the dismissal was a pointer, and `onCloseAutoFocus` preventDefaults on that path only. Keyboard dismissal still returns focus, which is the behaviour that matters for a11y.

**4. Menu surface matches `V2 Menu Dropdown`.** Checked against the reference code for node `7393:62008` in the product file rather than by eye. The panel takes `rounded-sm` (12px) and rows take `rounded-xs` (8px); padding, `backdrop-blur-[4px]` and the `0px 6px 12px rgba(0,0,0,0.1)` shadow all already matched. Checked rows stop being shaded.

The `sheet` path is deliberately 24px: it renders through `DrawerContent variant="menu"`, whose `MENU_CLASSES` is `rounded-lg`. A bottom sheet is an inset full-width surface with its own radius, not a scaled-up popper. Noted at the class so the two are not reconciled by mistake later.

## Review fixes

- Search field runs to the header padding. It was inset 16px left and 8px right, because the row kept a left-only inset from the layout that assumed a trailing close button; Figma's search row has no inset at all.
- Search icon is theme-aware (`content-primary`), at Jackson's request. Note this now differs from the placeholder beside it, which Figma draws muted.
- Two-line rows: avatar, count and icons sit on the title's line. The old `pt-1` was computed against the single-line title's 24px leading but only ever applied when a description is present, where the title is 14px/18px, so all four sat 3px low. Pre-existing on `main`.
- Dropped the inert `size` prop and `DropdownMenuCheckboxItemSize`.

## Not in scope, raising separately

Two pre-existing DropdownMenu deviations found while checking the above: `DropdownMenuRadioItem` is `px-4` where the design and `DropdownMenuCheckboxItem` are `px-3`, and the header divider is `pr-[4px]` in Figma against our full-bleed `-mx-2`. The second needs a decision rather than a fix, since the existing comment says the full bleed is deliberate.

**Lint note:** the 7 biome warnings in this file are pre-existing. `main` reports the identical 7 (1 complexity, 4 `useSortedClasses`, 2 `noNonNullAssertion`), so this PR adds none.

Full suite green, `tsc --noEmit` clean.

### DIFF
--- a/src/components/DropdownMenu/DropdownMenu.tsx
+++ b/src/components/DropdownMenu/DropdownMenu.tsx
@@ -9,6 +9,7 @@ import { IconButton } from "../IconButton/IconButton";
 import { CheckIcon } from "../Icons/CheckIcon";
 import { CloseIcon } from "../Icons/CloseIcon";
 import { SearchIcon } from "../Icons/SearchIcon";
+import { TickIcon } from "../Icons/TickIcon";
 
 // Movement, in CSS px, above which a touch press-and-release counts as a drag.
 const TAP_MOVEMENT_THRESHOLD_PX = 10;
@@ -217,20 +218,39 @@ export const DropdownMenuContent = React.forwardRef<
       // is enough to keep the menu flipping/shifting to stay on screen — no
       // hand-rolled reposition logic needed.
       collisionPadding = FLOATING_CONTENT_COLLISION_PADDING,
+      onPointerDownOutside,
+      onCloseAutoFocus,
       children,
       ...props
     },
     ref,
   ) => {
     const variant = React.useContext(DropdownMenuVariantContext);
+    // Radix returns focus to the trigger when the menu closes. That is right for
+    // the keyboard, but Chrome's :focus-visible heuristic also paints the ring on
+    // that programmatic refocus, so dismissing with a click left a focus ring
+    // sitting on the trigger. Track pointer dismissals and skip the restore for
+    // them only — Escape and Tab still hand focus back with the ring.
+    const dismissedByPointer = React.useRef(false);
 
     if (variant === "sheet") {
       return (
         <DrawerContent
           ref={ref}
           position="bottom"
-          variant="sheet"
-          className={cn("flex flex-col gap-1 p-1", className)}
+          // `"sheet"` sits flush to the bottom edge at full width; the design draws
+          // this panel inset from all three sides and rounded on every corner, which
+          // is `"menu"`. Both carry the same modal surface.
+          variant="menu"
+          // The design's `blur + shadow/menu` effect is a background blur, and the
+          // panel's own fill is opaque — so the blur has to go on the overlay to be
+          // visible at all. Scoped here rather than on DrawerOverlay so only menus
+          // shown this way blur the page behind them.
+          overlayProps={{ className: "backdrop-blur-[8px]" }}
+          // `pb-4` gives the design's 16px below the last row. The rows inset
+          // themselves horizontally (see the item's `mx-3`) rather than the panel
+          // padding doing it, so the header and its rule can still run full width.
+          className={cn("flex flex-col gap-1 p-1 pb-4", className)}
           style={style}
           {...props}
         >
@@ -243,10 +263,21 @@ export const DropdownMenuContent = React.forwardRef<
       <DropdownMenuPrimitive.Portal>
         <DropdownMenuPrimitive.Content
           ref={ref}
+          onPointerDownOutside={(event) => {
+            dismissedByPointer.current = true;
+            onPointerDownOutside?.(event);
+          }}
+          onCloseAutoFocus={(event) => {
+            if (dismissedByPointer.current) event.preventDefault();
+            dismissedByPointer.current = false;
+            onCloseAutoFocus?.(event);
+          }}
           sideOffset={sideOffset}
           collisionPadding={collisionPadding}
           className={cn(
-            "w-max min-w-(--radix-dropdown-menu-trigger-width) max-w-(--radix-dropdown-menu-content-available-width) overflow-y-auto rounded-sm border border-neutral-alphas-200 bg-surface-primary p-1 text-content-primary shadow-lg",
+            // `rounded-lg` is the panel radius `V2 Menu Dropdown` carries (24px).
+            // The 12px `rounded-sm` belongs to the rows inside it, not the panel.
+            "w-max min-w-(--radix-dropdown-menu-trigger-width) max-w-(--radix-dropdown-menu-content-available-width) overflow-y-auto rounded-lg border border-border-primary bg-surface-primary p-1 text-content-primary shadow-blur-menu backdrop-blur-[4px]",
             "data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
             "data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95",
             "data-[side=top]:slide-in-from-bottom-2 data-[side=bottom]:slide-in-from-top-2",
@@ -357,7 +388,14 @@ const ITEM_COUNT_TYPOGRAPHY: Record<"40" | "32", string> = {
 function SelectedCheckIndicator({ hasDescription }: { hasDescription: boolean }) {
   return (
     <CheckIcon
-      className={cn("size-4 shrink-0 text-content-primary", hasDescription && "self-start")}
+      className={cn(
+        "size-4 shrink-0 text-content-primary",
+        // The two-line layout switches the row to `items-start`, which would hang
+        // the tick off the title's line. A leading icon or avatar belongs there —
+        // it labels the title — but the tick is a property of the whole row, so it
+        // centres against both lines. {@link SelectItem} already does this.
+        hasDescription && "self-center",
+      )}
     />
   );
 }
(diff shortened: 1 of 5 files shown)


jacksonford-07:
Fixed in `e989cbd`. Panel goes `rounded-lg` (24px) to `rounded-sm` (12px).

I measured both panels in your screenshot rather than eyeballing it, because the two captures are at different zoom levels and that makes the corners misleading to compare by eye. Decoding the PNG and walking the panel silhouette against the page background:

- design panel: arc reaches the full left edge 12 rows below the top, panel 364px wide
- the other: 65 rows, panel 592px wide

The design one is self-consistent as a 1:1 capture of a ~364px dropdown at a 12px radius, which is what settled it. Also supporting: the variable set on the `V2 Menu Dropdown` documentation frame carries `Radius/rounded-sm: 12`, `rounded-xs: 8` and `rounded-2xs: 4`, and no `rounded-lg` at all.

Fixed the code comment too, which was wrong in a second way. It claimed the rows carry the 12px `rounded-sm` — they are `rounded-xs` (8px). So the old comment had the panel and the rows both a step too large.

Added a test asserting the panel is `rounded-sm` and neither `rounded-xs` nor `rounded-lg`, since there was no coverage on the radius, which is how it drifted in the first place. Split into separate negated assertions, per your note on #623.

One caveat on how I got there: `V2 Menu Dropdown` was updated on 30 Jul, and I could not open it to read the radius directly — search returns a component key without a node id, and this file exposes only one page to the MCP. So the 12px is derived from the measurement plus the token scale, not read off the node. If you have the node link, worth a second look.

97 tests, `tsc --noEmit` clean.

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -243,10 +263,21 @@ export const DropdownMenuContent = React.forwardRef<
...
+          onCloseAutoFocus={(event) => {
+            if (dismissedByPointer.current) event.preventDefault();
+            dismissedByPointer.current = false;
+            onCloseAutoFocus?.(event);
+          }}
           sideOffset={sideOffset}
           collisionPadding={collisionPadding}
           className={cn(
-            "w-max min-w-(--radix-dropdown-menu-trigger-width) max-w-(--radix-dropdown-menu-content-available-width) overflow-y-auto rounded-sm border border-neutral-alphas-200 bg-surface-primary p-1 text-content-primary shadow-lg",
+            // `rounded-sm` (12px) is the panel radius `V2 Menu Dropdown` carries.
```
alfonsovo-fv: The description and this comment disagree, and they disagree about both tokens. The description says "the panel takes `rounded-lg` (24px)" and that "the 12px `rounded-sm` belongs to the rows"; the code applies `rounded-sm` to the panel and says `rounded-xs` (8px) belongs to the rows.

Worth resolving rather than leaving, because the sheet path is genuinely 24px: it goes through `DrawerContent variant="menu"`, whose `MENU_CLASSES` is `rounded-lg`. So the same menu is 12px as a popper and 24px as a sheet. That may well be deliberate, but as it stands the description describes one and the comment the other, and someone reconciling them later has no way to tell which was intended.

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -882,3 +977,84 @@ export const DropdownMenuRadioItem = React.forwardRef<
...
+ *   avatar={<Avatar size={32} src={creator.avatarUrl} />}
+ * >
+ *   @sofiabloom
+ * </DropdownMenuCheckboxItem>
+ * ```
+ */
+export const DropdownMenuCheckboxItem = React.forwardRef<
+  React.ComponentRef<typeof DropdownMenuPrimitive.CheckboxItem>,
+  DropdownMenuCheckboxItemProps
+>(({ className, children, helper, avatar, size: _size = "40", ...props }, ref) => {
```
alfonsovo-fv: `size` does nothing. It is destructured as `_size`, never read, and `DropdownMenuCheckboxItemSize` is a single-member union exported from `src/index.ts` alongside it, so this ships as documented public API with a `@default` that has no effect.

I can see `DropdownMenuRadioItem` on line 935 already does exactly this, so it is a pattern rather than an oversight. But that one is at least pre-existing, whereas this is a new export, and an inert prop is worse on a new component than an old one: a consumer sets `size="40"`, nothing happens, and there is no way to tell whether that is because 40 is the default or because the prop is ignored.

Either wire it up or drop both the prop and the type until there is a second size. Worth a ticket for the RadioItem one either way.

alfonsovo-fv (commented):
Reviewed as the four concerns you split it into. The `shadow-blur-menu` change is a genuine improvement on what I flagged in #625: replacing `shadow-lg` rather than stacking on top of it means tailwind-merge no longer has two shadows it cannot dedupe.

Two things, one of which is just the description being out of step with the code.

This also restyles the existing menu for every current consumer: panel border, shadow and blur, header title weight, the full-bleed rule, the search field surface and icon colour, the two-line title type, and the tick moving to `self-center`. Only one file in `src/components` uses `DropdownMenu`, so internally it is contained, but published consumers get all of it on a minor. Same treatment as #627 please, a line in the description so it is discoverable without diffing.

src/components/DropdownMenu/DropdownMenu.tsx
jacksonford-07: Both wrong, and the description was the wrong one. Resolved in favour of the code.

I had already corrected the comment before your review landed, after Jackson flagged the panel radius against Figma. The authoritative value is `rounded-sm` (12px), read off the reference code for `V2 Menu Dropdown` in the product file, node `7393:62008`:

```
rounded-[var(--radius/rounded-sm,12px)]
```

Rows are `rounded-[var(--radius/rounded-xs,8px)]`, so the current comment is right on both tokens and the description was wrong on both. Description updated.

Worth saying the description was not just stale — I wrote it asserting a Figma value I had not actually read, and it took Jackson's eye plus a proper node read to settle it. Same failure mode as the two comment findings on #635.

On the sheet: verified, and it is deliberate. `DrawerContent variant="menu"` carries `MENU_CLASSES` with `rounded-lg`, and the sheet branch does pass `variant="menu"`. A bottom sheet is an inset full-width surface with its own radius rather than a scaled-up popper. Added that to the comment with the token path, since as you say there was no way to tell which was intended.

Everything else in the panel checks out against that node: padding 4px, `backdrop-blur-[4px]`, `0px 6px 12px rgba(0,0,0,0.1)`, search input `rounded-xs` with `px-3 py-1`, checkbox `rounded-2xs`.

src/components/DropdownMenu/DropdownMenu.tsx
jacksonford-07: Dropped, per your first option. `size` and `DropdownMenuCheckboxItemSize` are both gone, along with the `src/index.ts` export. Nothing is released yet so there is no compatibility cost.

Took the "either wire it up or drop it" as drop because there is no second height to wire it to — a one-member union is a prop that can only ever be set to its default.

Agreed on the asymmetry you drew: `DropdownMenuRadioItem` doing the same thing is pre-existing and shipped, this was new, and an inert prop on a new export is worse because a consumer has no way to tell "40 is the default" from "the prop is ignored".

Leaving `DropdownMenuRadioItem` alone in this PR and will raise it. Two other pre-existing DropdownMenu items for the same ticket, both found while chasing Jackson's screenshots:

- `DropdownMenuRadioItem` is `px-4` where the design and our own `DropdownMenuCheckboxItem` are `px-3`, so radio rows indent 4px further than checkbox rows in the same menu
- the header's `V2 Divider` is `pr-[4px]` in Figma against our full-bleed `-mx-2`, which has a comment claiming the rule spans the panel deliberately, so that one needs a decision rather than a fix

jacksonford-07:
Follow-up raised as #640 with the three pre-existing deviations found while reviewing this one: the item row radius (8px, design is 12px), the radio item padding (`px-4`, design is `px-3`) and the count typography at size 40 (16px, design is 14px). Each is commented at the line with its Figma node id and has a test.

The header divider is deliberately not in it. The evidence is inconsistent across header variants, so it needs a design answer rather than a change — detail in #640's description.

jacksonford-07:
Correcting something I claimed earlier in this PR, found by rendering it locally rather than reading classes.

**I said the two-line misalignment was pre-existing on `main`. It is not — this PR introduced it.**

The `pt-1` and its comment are on `main`, which is what I checked, and that is why I said so. But on `main` the two-line title has no typography of its own: it is `<span className="truncate">` inheriting the row's 16px/24px. A 16px icon at `pt-1` centres on a 24px line exactly, so `main` is correct and its comment is accurate for the code it sits in.

This PR takes the two-line title to `typography-body-small-14px-semibold` (14px/18px), matching `V2 Menu Item` Type=Description. That is the right change, but it moved the title's centre from 20 to 17 and left the `pt-1` behind, so every sibling went 3px low. Self-inflicted, in this diff.

Measured on `main` + nothing else, `FeatureItems` dark:

```
row                    title       titleC   iconC
Edit profile (2 line)  16px/24px   20       24    <- 24px icon, 4px low
Alex Smith  (2 line)   16px/24px   20       -
```

So `main` does have a narrower version of the same bug — a **24px** icon sits 4px low even there, because `pt-1` only centres a 16px one. Figma's `V2 Menu Item` uses 16px icons throughout, and our own stories pass 24px ones. That part is genuinely pre-existing.

Verified the fix on this branch, same story:

```
row                    title       titleC   iconC   countC
Jane Doe    (1 line)   16px/24px   20.8     20.8    -
Alex Smith  (2 line)   14px/18px   17       -       17
Favourites  (1 line)   16px/24px   20       20      20
Edit profile(2 line)   14px/18px   17       17      -
```

The `h-[18px]` box holds for any sibling size, which is why the 24px icon on row 4 now lands on 17 rather than needing its own offset.

Two other things the render turned up, both outside this PR:

- `main`'s panel is already `rounded-sm` (12px). The `rounded-lg` I "fixed" here came from the insights batch, so that was also this branch's own regression rather than a pre-existing one. The value is right either way; my description overstated what it was fixing, and it is now corrected.
- `Size32` rows are 38px, not 32px, when given a 24px `leadingIcon` (7 + 24 + 7). The component already tightens `py` for a 24px **avatar** at size 32 and has a comment saying why, but does nothing for a 24px icon. Pre-existing, raising separately.

alfonsovo-fv (approved):
Both resolved, approving.

The radius is now consistent between the code, the comment and the description, and the description explains why the sheet is deliberately 24px through `DrawerContent variant="menu"` rather than leaving the two to look like a contradiction. That is the useful outcome, since the risk was someone reconciling them later by guessing.

Inert `size` prop and its single-member type are both gone from the component and from `src/index.ts`. Leaving the pre-existing `DropdownMenuRadioItem` one for a ticket is fine.

PR #640, opened 2026-07-31
fix(DropdownMenu): align row radius, radio padding and count size to V2

jacksonford-07:
Follow-up to #639. Three pre-existing deviations from the design, all on `main` today and none introduced by that PR. Split out because each one changes every dropdown in eden and fadmin, not just the rows #639 touches.

Every value here is read from the Figma reference code, not inferred from a screenshot, and each is commented at the line with its node id — the recurring failure on this component has been comments asserting values nobody checked.

| what | was | is | source |
| --- | --- | --- | --- |
| `DropdownMenuItem` row radius | `rounded-xs` (8px) | `rounded-sm` (12px) | `V2 Menu Item`, node `21542:8629` |
| `DropdownMenuRadioItem` padding | `px-4` (16px) | `px-3` (12px) | `V2 Menu Radio Item`, node `7393:62008` |
| count typography at size 40 | 16px/24px | 14px/18px | `V2 Menu Item`, node `21542:8629` |

**The two row radii genuinely differ, and that is not a mistake.** `V2 Menu Item` is `rounded-sm` (12px); `V2 Menu Radio Item` is `rounded-xs` (8px). So this changes the plain item and deliberately leaves the radio and checkbox items alone. There is a test asserting the radio item *stays* at 8px, so the next person to notice the inconsistency finds the reason instead of "fixing" it.

**The radio padding is also a self-consistency fix.** At `px-4` a radio row indented 4px further than a `DropdownMenuCheckboxItem` in the same menu.

**The count size compounded the alignment fix in #639.** At 16px its line box is 24px against the two-line layout's 18px title leading, so it was both the wrong size and the wrong height. #639 centred it regardless, which had the effect of hiding the size error.

## Tests

Four, one per change plus the radio-radius guard. Verified against `main`: the three change assertions fail, the guard passes on both.

```
× gives a menu item the 12px row radius
✓ keeps the 8px radius on a radio item, which the design draws differently
× indents a radio item the same as any other row
× draws the count at 14px at both row heights
```

Full suite 4530, `tsc --noEmit` clean.

## Not included, needs a decision

The header divider. Ours is full-bleed `-mx-2` with a comment saying the rule spans the panel deliberately; the search header in node `7393:62008` draws its `V2 Divider` with `pr-[4px]`, inset from both edges. I did not change it, because the evidence is inconsistent across variants — the Default Desktop header (node `21549:6769`) has **no divider at all** and uses 12px padding where that instance uses 4px. Reversing a documented deliberate choice on one node needs a design answer, not a guess.

Worth flagging for the same conversation: the search instance's 4px header padding and the library Default variant's 12px disagree, so one of them is probably not the variant we should be matching.

### DIFF
--- a/src/components/DropdownMenu/DropdownMenu.test.tsx
+++ b/src/components/DropdownMenu/DropdownMenu.test.tsx
@@ -83,6 +83,53 @@ describe("DropdownMenu", () => {
       expect(panel).not.toHaveClass("rounded-lg");
     });
 
+    // Each of these three drifted from the design unnoticed, so they are asserted
+    // rather than left to the eye. Node references are in the component.
+    it("gives a menu item the 12px row radius", () => {
+      renderMenu(<DropdownMenuItem>Item</DropdownMenuItem>);
+      const item = screen.getByRole("menuitem");
+      expect(item).toHaveClass("rounded-sm");
+      expect(item).not.toHaveClass("rounded-xs");
+    });
+
+    it("keeps the 8px radius on a radio item, which the design draws differently", () => {
+      renderMenu(
+        <DropdownMenuRadioGroup value="a">
+          <DropdownMenuRadioItem value="a">A</DropdownMenuRadioItem>
+        </DropdownMenuRadioGroup>,
+      );
+      const item = screen.getByRole("menuitemradio");
+      expect(item).toHaveClass("rounded-xs");
+      expect(item).not.toHaveClass("rounded-sm");
+    });
+
+    it("indents a radio item the same as any other row", () => {
+      renderMenu(
+        <DropdownMenuRadioGroup value="a">
+          <DropdownMenuRadioItem value="a">A</DropdownMenuRadioItem>
+        </DropdownMenuRadioGroup>,
+      );
+      const item = screen.getByRole("menuitemradio");
+      expect(item).toHaveClass("px-3");
+      expect(item).not.toHaveClass("px-4");
+    });
+
+    it("draws the count at 14px at both row heights", () => {
+      renderMenu(
+        <>
+          <DropdownMenuItem size="40" count="12">
+            Forty
+          </DropdownMenuItem>
+          <DropdownMenuItem size="32" count="34">
+            Thirty two
+          </DropdownMenuItem>
+        </>,
+      );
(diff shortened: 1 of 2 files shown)


jacksonford-07:
Restacked onto #639's branch and verified in a local Storybook rather than by reading class names. That check found a real problem with how I had based this PR.

**Why it had to move.** This was branched off `main`, where the two-line title inherits the row's 16px/24px. Dropping the count to 14px/18px there puts it 3px *above* the title. Measured on the old base:

```
two-line row:  title 16px/24px centre 20   count 14px/18px centre 17
```

The count change is only correct once #639 lands, because that is what takes the two-line title to 14px/18px and wraps the siblings in an 18px box. So #640 could not merge first without introducing a misalignment. Base is now `ds/dropdown-multiselect`; it will retarget to `main` when #639 merges.

**Verified on the combined branch**, `FeatureItems` dark:

```
row                     rowRadius  title        titleC  iconC  count        countC
Jane Doe (1 line)       12px       16px/24px    20.8    20.8   -            -
Alex Smith (2 line)     12px       14px/18px    17      -      14px/18px    17
Favourites (1 line)     12px       16px/24px    20      20     14px/18px    20
Edit profile (2 line)   12px       14px/18px    17      17     -            -
```

Every row's siblings share their title's centre, and the count is 14px at both heights. Panel 12px, rows 12px.

Radio item, `WithRadioGroup` dark: radius 8px, padding 12px both sides, indicator centre 20 against title line centre 20. So the radio keeps the 8px the design gives it and now indents the same as every other row.

103 tests, `tsc --noEmit` clean.

alfonsovo-fv (approved):
Approving, nothing to flag. Citing the Figma node id at each changed line is the right response to a component whose recurring problem has been comments asserting unverified values, and stating that the three assertions fail against `main` while the guard passes on both is what makes the tests worth having.

The radio-radius guard is the best part: an intentional inconsistency that nobody documented is exactly what gets "tidied up" by the next person, and now it cannot be without a test going red.

Also the right call to leave the header divider alone. Reversing a documented deliberate choice on the strength of one node, when a sibling variant of the same component disagrees with it, does need a design answer rather than a guess.

PR #641, opened 2026-07-31
feat(CountryFlag, Demographics): add the country breakdown row and its flags

jacksonford-07:
Seventh of the `insights-components` extraction, and the last blocker for eden's agency insights. This is the whole country-breakdown feature: the row, the flags, and the two `ProgressBar` additions the row needs.

`CountryFlag` renders one of 265 circular flags. Used by the top-countries breakdown, where the code is a runtime value from the API, so the set cannot be subset at build time.

## What is here

- `CountryFlag` and the `./flags` entry point
- `Demographics`, the breakdown row
- `ProgressBar` gains `size="medium"` (8px) and `variant="sky"`
- `special-chart-muted-sky` token, `theme.css` regenerated via `node src/styles/buildStyles.js`

The last three are what `tsc` demanded once `Demographics` landed: it needs the sky `ProgressBar`, which needs the muted-sky token. That token is the first from the `Color/Special/Chart/Muted/*` ramp I flagged as unexported in the contrast thread on #629, and it answers that question — the muted ramp is the **track** behind a full-strength bar, not an alternative to it.

`Demographics` does **not** depend on `CountryFlag`; it takes `icon` as a plain `ReactNode`. So the generic row stays cheap and only the flag consumer pays for the flags.

## On the 4,298 lines

Jackson asked whether this could be smaller. I measured it rather than guessed, and the count is almost all generated data.

**What the lines are.** 3,650 of 4,298 are `flagShapes.ts`, carrying an `AUTO-GENERATED - DO NOT MODIFY` header and regenerated by `pnpm flags:generate` from `circle-flags@2.8.3` (MIT). The generator is included so the data is reproducible rather than a one-off paste.

**What it costs, from the build:**

```
dist/flags.mjs                              0.15 kB │ gzip:  0.13 kB
dist/components/CountryFlag/flagShapes.mjs  163.20 kB │ gzip: 38.00 kB
```

**It is isolated, which is the part that matters.** `flagShapes` compiles to its own chunk, and only `CountryFlag.mjs` imports it. Checked against the built output: `dist/index.mjs` contains zero references to `flagShapes`, so nothing reachable from the root `@fanvue/ui` entry can pull it in. Only a file importing `@fanvue/ui/flags` pays the 38 kB, which in eden is the one card that draws flags.

**What I considered and rejected.** Splitting into 265 modules would tree-shake, but the country is runtime-determined so consumers would need a dynamic import per flag and `dist` would gain 265 files. Lazy-loading the whole map inside the component would move 38 kB off the initial chunk but makes a small presentational component async. Neither is worth it while the cost is already confined to one subpath.

## Verified in the showcase, not just in tests

`AllFlags` renders every entry, which is the real check on generated data:

```
rendered: 265      withTitle: 265     emptyArtwork: 0
zeroSized: 0       maskIdsUnique: true (265/265)   colons in mask ids: none
```

`Sizes` renders at 16/20/24/32 with `viewBox="0 0 512 512"` and correct box sizes.

**Unmapped codes render the source's own `xx` placeholder disc, not nothing.** `circle-flags` ships `xx` as a plain grey circle, so it goes through the identical path as a real flag and inherits the same 512 viewBox and circular mask. A hand-drawn `bg-content-tertiary` span would be the one glyph in the set that is not real artwork. Two tests assert it: that the fallback draws on the same viewBox with shapes present, and that an unmapped code renders identically to an explicit `xx`.

From the `Demographics` showcase demo, four real codes plus `zz`:

```
US #d80027   GB #0052b4   NL #a2001d   zz #acabb1 (grey placeholder)
bar: track rgb(227,243,254)  fill rgb(79,178,249)  h 8px
```

Each flag instance takes a `useId`-derived mask id with the colons stripped, since a `url(#...)` reference cannot carry them — same fix as #637.

**a11y.** The flag is decorative by default (`aria-hidden`), and `role="img"` with a `<title>` when `label` is given; that is the caller's choice because only the host can translate a country name. `Demographics` names its bar after the row, so a stacked breakdown announces the country rather than ten bars each saying "Progress" and a number. `label` is typed `string` rather than `ReactNode` for that reason — see the thread below. Verified across all five demo rows, including the one with no icon:

```
United States 52.6 · United Kingdom 21.4 · Netherlands 12.8 · Unknown 8.2 · Direct 64
```

## Public API

Adds a `./flags` subpath to `package.json` exports plus the matching `vite.config.ts` entry. That is the piece worth your attention rather than the path data — it is a new public entry point, deliberately separate from the root so the flag payload stays opt-in.

Full suite 4648, `tsc --noEmit` clean, `pnpm build` clean.


jacksonford-07:
Changed the unknown-code behaviour on Jackson's suggestion. It now renders a grey placeholder disc rather than nothing.

Went with the source's own `xx` artwork instead of drawing a disc from a token. `circle-flags` already ships it, and inspecting the data it is exactly what was wanted:

```
xx: 3 shapes [path, circle, path]  fills = #acabb1, #eee
```

So it is a plain grey circle. Three reasons to prefer it over a `bg-content-tertiary` span:

- it costs nothing, being one of the 265 already in the chunk
- it renders through the identical path, so it inherits the same 512 viewBox and circular mask as a real flag; a hand-drawn disc would be the one glyph in the set that is not real artwork
- `xx` is the source's documented placeholder, so it stays correct if the artwork is regenerated

**Verified rendering**, `Unknown country code` story, a real flag beside an unmapped code:

```
NL   20x20  title "Netherlands"  fills #fff #eee #a2001d #0052b4
zz   20x20  title "Unknown"      fills #fff #eee #acabb1 #acabb1
```

Same box, same treatment, grey.

One thing worth saying since it was the reason I checked before agreeing: this was **not** an alignment fix. `Demographics` reserves the slot regardless, because `icon` is a truthy JSX element even when `CountryFlag` returns `null`, so the `size-5` wrapper still renders and rows stayed aligned either way. The change is purely that a blank gap where every other row has a glyph reads as a broken image rather than an unknown country.

Replaced the test that asserted the old behaviour with two: one that the fallback draws on the same viewBox with shapes present, and one that an unmapped code renders identically to an explicit `xx`. The second is the one that would catch someone changing the placeholder to something hand-rolled.

Full suite 4572.

jacksonford-07:
Merged #642 into this PR and restored the real flags to the `Demographics` stories, so this is now the whole country-breakdown feature in one place. Renamed to match.

Splitting them was my call, not asked for, and it was the wrong one for two reasons Jackson pointed out: two stacked PRs produce two Chromatic builds covering overlapping snapshots, and keeping `Demographics` independent meant its stories could not render a real flag — which is the one thing the component exists to do.

**What is now here**

- `CountryFlag` and the `./flags` entry point
- `Demographics`, the breakdown row
- `ProgressBar` gains `size="medium"` (8px) and `variant="sky"`
- `special-chart-muted-sky` token, `theme.css` regenerated via `node src/styles/buildStyles.js`

The three after the first are what `tsc` demanded once `Demographics` landed — it needs the sky ProgressBar, which needs the muted-sky token. That token is the first from the `Color/Special/Chart/Muted/*` ramp I flagged as unexported in the contrast thread on #629, and it answers that question: the muted ramp is the **track** behind a full-strength bar, not an alternative to it.

**On whether the flags sit over anything** — they do not, and the DOM says so rather than me. `Demographics` renders the slot as a single wrapper with one child and no background:

```
slotChildren: 1
slotBg:       rgba(0, 0, 0, 0)
```

So whatever is passed as `icon` *is* the glyph. `Stacked breakdown` dark, five real flags, each the sole child of a 20px transparent slot:

```
US #d80027   GB #0052b4   NL #a2001d   IT #6da544   ES #ffda44
bar: track rgb(48,80,106)  fill rgb(79,178,249)  h 8px
```

**`WithoutIcon` checked too**, since that was the other half of the question: with no `icon` the `size-5` slot is not rendered at all, so labels sit flush at x=16 exactly aligned with their bars rather than leaving an orphaned gap.

Full suite 4580, `tsc --noEmit` clean.

alfonsovo-fv (commented):
The component itself is good and the bundle argument is sound. Two things, one of which affects how this merges.

**This and #642 overlap, and both target `main`.** `git diff` between the two branches shows #642 is contained in this one apart from `Demographics.stories.tsx`: `Demographics.tsx`, `ProgressBar.tsx`'s `sky` and `medium` additions, the `muted-sky` token in `styleTokens.json` and all three `theme.css` blocks are in both. Your description says Demographics "stacks on this", but its base is `main`, not `ds/country-flag`. So whichever lands first, the other conflicts on the token files. Retarget #642 onto this branch, or drop the Demographics files from here.

**The description contradicts the code on unmapped codes.** The verification section says "`Unknown country code` renders nothing, as an unmapped code returns `null`", but line 96 falls back to the `xx` placeholder disc, the comment explains why rendering nothing would be wrong, and two tests assert the placeholder. The code is right and the reasoning is right; it is the description that is stale, and it is the thing a reviewer reads first.

Two things I checked that are fine, so you do not need to revisit them. The `replace(/:/g, "")` is load-bearing despite this repo dev-testing on React 19 — 19.2.3 returns `_r_0_` with no colons, but `peerDependencies` allows `^18`, where it is `:r0:`. And your isolation invariant is enforced rather than just verified once: if `flagShapes` ever leaked into the root entry, 193 kB plus 163 kB would blow the existing 220 kB `dist/index.mjs` budget immediately. The flags chunk itself has no budget, so a regenerate against a newer `circle-flags` could grow unnoticed, but that is a suggestion rather than a problem.

src/components/Demographics/Demographics.tsx
```diff
@@ -0,0 +1,57 @@
...
+        </span>
+        <span className="typography-body-small-14px-regular shrink-0 text-content-primary">
+          {formattedValue}
+        </span>
+      </div>
+      <ProgressBar
+        variant="sky"
+        size="medium"
+        value={value}
+        ariaLabel={typeof label === "string" ? label : undefined}
```
alfonsovo-fv: Re-raising from #642, since closing it took the thread with it and the code landed here unchanged.

This silently drops the bar's accessible name for anything that is not a string. `label` is typed `React.ReactNode`, so a translated element or a fragment makes this `undefined`, and `ProgressBar` then falls back to `aria-label={ariaLabel ?? "Progress"}` (line 162). A stacked breakdown becomes ten bars each announcing "Progress" and a number, with nothing saying which country.

The root of it is that the doc says "Pass a translated string for i18n" while the type allows any node, so the type permits the path that breaks. Narrowing `label` to `string` matches what the doc already asks for and is the cheaper of the two fixes; pointing at the rendered label with `aria-labelledby` also works but needs `ProgressBar` to forward it.

src/components/CountryFlag/CountryFlag.tsx
```diff
@@ -0,0 +1,118 @@
...
+    // can't carry them.
+    const maskId = `country-flag-${generatedId.replace(/:/g, "")}`;
+
+    // Unmapped codes fall back to the source's own `xx` placeholder — a plain
+    // grey disc — rather than rendering nothing. A live feed can hand us a code
+    // we hold no artwork for, and a blank where every other row has a glyph
+    // reads as a broken image rather than an unknown country. Using `xx` keeps
+    // the placeholder on the same viewBox and circular mask as a real flag,
+    // which a hand-drawn disc would not.
+    const shapes =
```
alfonsovo-fv, line 96: Still worth a description edit: the verification section says "`Unknown country code` renders nothing, as an unmapped code returns `null`", but this falls back to the `xx` placeholder disc, the comment here explains why rendering nothing would be wrong, and two tests assert the placeholder.

The code and the reasoning are both right, it is only the description that is stale — and after the radius mix-up on #639 it is worth being strict about it, because the description is what a reviewer trusts before reading the diff.

alfonsovo-fv (commented):
Merging #642 into this rather than retargeting it resolves the overlap cleanly, so that is sorted.

Two things carried over that are still open, one of which lost its thread when #642 closed, so re-raising it here rather than letting it disappear.

src/components/Demographics/Demographics.tsx
jacksonford-07: Fixed by narrowing `label` to `string`, which was your cheaper option and the one that matches what the doc already asked for.

You were right that the type permitted the path that breaks. `ariaLabel={typeof label === "string" ? label : undefined}` was defensive code guarding against a case the props should never have allowed in the first place, so removing the ability to pass a node removes the need for the guard:

```tsx
<ProgressBar variant="sky" size="medium" value={value} ariaLabel={label} />
```

The doc comment now carries the reason, so the next person does not widen it back:

```
A string rather than a node because it is also the bar's accessible name:
a node would leave the bar announcing "Progress" and a bare number, with
nothing saying which row it belongs to.
```

`aria-labelledby` was the other option and I did not take it: it needs `ProgressBar` to forward the attribute and an id on the label span, which is more surface for a component whose label is a country name in every real use.

No consumer breakage — every call site in stories, tests and the new showcase demo already passed a string, and `tsc --noEmit` is clean.

**Verified in the running showcase** rather than only in jsdom, all five rows including the one with no icon:

```
United States 52.6 · United Kingdom 21.4 · Netherlands 12.8 · Unknown 8.2 · Direct 64
```

Five bars, five distinct accessible names, no "Progress". Full suite 4648.

src/components/CountryFlag/CountryFlag.tsx
jacksonford-07: Description rewritten. You were right to keep pushing on this — the stale line was not the only one, since merging #642 in made most of the framing wrong.

What was wrong and is now fixed:

- "`Unknown country code` renders nothing, as an unmapped code returns `null`" → now states the `xx` placeholder, why the source's own artwork beats a hand-drawn disc, and which two tests hold it
- "the last blocker other than `Demographics`, which stacks on this" → `Demographics` is in this PR now, so it listed as the second thing here
- 3,890 lines → 4,298, with the generated share restated (3,650 of them)
- full suite 4571 → 4648
- added the `ProgressBar` and `muted-sky` additions, which were not mentioned at all despite being public API changes

Taking the point about description trust generally, not just here. The radius mix-up on #639 came from me asserting a "pre-existing on main" claim I had not checked, and a stale description is the same failure one step earlier. So the a11y section now quotes the five accessible names read out of the live DOM rather than describing intent.

jacksonford-07:
Both threads addressed and resolved, plus one thing you did not flag that I found while checking.

**`label` narrowed to `string`** — your cheaper fix. The `typeof label === "string"` guard is gone with it, since the props no longer allow the case it was guarding.

**Description rewritten.** The unmapped-code line was stale, and so was most of the rest once #642 was merged in: the "stacks on this" framing, the line count, the suite count, and the `ProgressBar` / `muted-sky` additions that were not mentioned at all despite being public API.

**Added the `Demographics` showcase demo.** `AGENTS.md` line 71 asks for an `App.tsx` demo with any new component and this PR had none for either of them, which I missed. It doubles as the first showcase use of the `./flags` subpath, so the entry point is exercised outside its own stories.

Also rebased onto `main` for #643, which is why Chromatic has been bounced. No conflicts, and 4648 tests pass on the new base.

**Verified in the running showcase** rather than only in jsdom, which is what caught that the placeholder needed to be checked in the same DOM as the real flags:

```
US #d80027   GB #0052b4   NL #a2001d   zz #acabb1 (grey placeholder)
bar: track rgb(227,243,254)  fill rgb(79,178,249)  h 8px
accessible names: United States · United Kingdom · Netherlands · Unknown · Direct
label x: 268 268 268 268 240    bar x: 240 240 240 240 240
```

The last line is the icon-slot question from earlier, measured: the four rows with a flag inset their label by 28px (20px slot + 8px gap), and the `Direct` row with no icon sits flush at 240 with its own bar rather than leaving an orphaned gap.

On your flags-chunk suggestion — you are right that a regenerate against a newer `circle-flags` could grow unnoticed since the chunk has no `size-limit` budget. Not adding one here because it needs a number chosen against the current 38 kB rather than picked now, but it is worth a follow-up and I will raise it.

alfonsovo-fv (approved):

PR #642, opened 2026-07-31
feat(Demographics): add the breakdown row, with its sky ProgressBar and token

jacksonford-07:
Eighth and last of the `insights-components` extraction. Stacked on #641 because the stories render real flags; it will retarget to `main` when that merges. After this, eden has every symbol it imports.

One row of a breakdown: leading glyph, label, share at the far end, bar underneath. Stacked for the top-countries card on agency insights.

## Three changes, not one

I planned this as a 168-line component PR. It is not — `tsc` refused it, which is how I found the rest. Reviewing them separately:

**1. `special-chart-muted-sky` token.** `#e3f3fe` light, `#30506a` dark. Added to `styleTokens.json` and `theme.css` regenerated via `node src/styles/buildStyles.js` rather than hand-edited — the file carries an `AUTO-GENERATED` header. The regenerated output differs from `main` by exactly the 3 expected lines.

This is the first token from the `Color/Special/Chart/Muted/*` ramp, which is the ramp I flagged as entirely unexported in the contrast thread on #629. So that question now has one concrete answer: the muted ramp is the **track** colour behind a full-strength chart bar, not an alternative to it.

**2. `ProgressBar` gains `size="medium"` (8px) and `variant="sky"`.** `sky` is `bg-special-chart-sky` on a `bg-special-chart-muted-sky` track. Additive — every existing variant and size is untouched, which the 48 existing ProgressBar tests confirm.

**3. `Demographics` itself.** Takes `icon` as a plain `ReactNode`, so it has no dependency on `CountryFlag`; only its stories do. That keeps the generic row cheap and the 38 kB flag payload confined to `@fanvue/ui/flags` (see #641).

## Verified in Storybook, both themes

The token is the part that could have failed silently — a missing custom property yields no background rather than an error, so a track would simply be invisible. Measured on `Breakdown`:

```
light   token #e3f3feff   track rgb(227,243,254)   fill rgb(79,178,249)   trackH 8px
dark    token #30506aff   track rgb(48,80,106)     fill rgb(79,178,249)   label #fff
```

Five rows with `aria-valuenow` of 52.6 / 29.2 / 7.6 / 7.3 / 6.9 and fill widths proportional to each.

Full suite 4579, `tsc --noEmit` clean.

### DIFF
--- a/src/components/Demographics/Demographics.tsx
+++ b/src/components/Demographics/Demographics.tsx
@@ -1,0 +1,57 @@
+import * as React from "react";
+import { cn } from "../../utils/cn";
+import { ProgressBar } from "../ProgressBar/ProgressBar";
+
+/** Props for {@link Demographics}. */
+export interface DemographicsProps extends Omit<React.HTMLAttributes<HTMLDivElement>, "title"> {
+  /** Row label, e.g. a country name. Pass a translated string for i18n. */
+  label: React.ReactNode;
+  /** Share of the total, 0–100. Drives the bar and is announced to assistive tech. */
+  value: number;
+  /** Formatted share shown at the end of the row, e.g. `"52.6%"`. */
+  formattedValue: React.ReactNode;
+  /** Leading 20px glyph — a flag for a country row. */
+  icon?: React.ReactNode;
+}
+
+/**
+ * One row of a breakdown: a leading glyph and label, its share at the far end,
+ * and a bar underneath. Stack these to show how a total splits across
+ * categories — the top-countries breakdown on the insights dashboard, and the
+ * same list expanded inside a dialog.
+ *
+ * @example
+ * ```tsx
+ * <Demographics
+ *   icon={<span aria-hidden>🇺🇸</span>}
+ *   label="United States"
+ *   value={52.6}
+ *   formattedValue="52.6%"
+ * />
+ * ```
+ */
+export const Demographics = React.forwardRef<HTMLDivElement, DemographicsProps>(
+  ({ label, value, formattedValue, icon, className, ...props }, ref) => (
+    <div ref={ref} className={cn("flex w-full flex-col gap-2", className)} {...props}>
+      <div className="flex w-full items-center justify-between gap-2">
+        <span className="flex min-w-0 items-center gap-2">
+          {icon && <span className="flex size-5 shrink-0 items-center justify-center">{icon}</span>}
+          <span className="typography-body-small-14px-semibold truncate text-content-primary">
+            {label}
+          </span>
+        </span>
+        <span className="typography-body-small-14px-regular shrink-0 text-content-primary">
+          {formattedValue}
+        </span>
+      </div>
+      <ProgressBar
+        variant="sky"
+        size="medium"
+        value={value}
+        ariaLabel={typeof label === "string" ? label : undefined}
+      />
+    </div>
+  ),
+);
+
+Demographics.displayName = "Demographics";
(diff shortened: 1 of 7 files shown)


jacksonford-07:
Now independent of #641 and based on `main`, so either can merge first in any order.

The only thing coupling them was `Demographics.stories.tsx` importing `CountryFlag`. The component itself never did — `icon` is a plain `ReactNode`. So the stories now fill the slot with a local disc:

```tsx
const Glyph = () => <span aria-hidden className="size-5 rounded-full bg-special-chart-sky" />;
```

That is arguably the better story for a generic row anyway: it shows the slot takes anything, rather than implying a dependency on flags. Agency insights pairs it with a real `CountryFlag` from `@fanvue/ui/flags`, which the comment above it says.

**Zero shared files with #641**, checked rather than assumed:

```
#641  package.json, vite.config.ts, scripts/generate-flags.mjs,
      src/flags.ts, src/components/CountryFlag/*

#642  src/components/Demographics/*, src/components/ProgressBar/ProgressBar.tsx,
      src/index.ts, src/styles/styleTokens.json, src/styles/theme.css

overlap: none
```

Re-verified rendering after the decoupling, `Stacked breakdown` dark: 5 rows, 20px glyphs, track `rgb(48,80,106)` at 8px from `--color-special-chart-muted-sky` `#30506aff`, fill `rgb(79,178,249)`, `aria-valuenow` 52.6 on the first row.

Full suite 4567, `tsc --noEmit` clean.

jacksonford-07:
Folded into #641. Splitting these produced two Chromatic builds covering overlapping snapshots, and keeping them apart meant the Demographics stories could not show real flags — which is the whole point of the component. #641 now carries both, with the countries restored.

Verified in Storybook, `Stacked breakdown` dark: five real flags (US `#d80027`, GB `#0052b4`, NL `#a2001d`, IT `#6da544`, ES `#ffda44`), each the single child of a transparent 20px slot. `WithoutIcon` renders no slot at all, so labels sit flush at x=16 aligned with their bars.

src/components/Demographics/Demographics.tsx
```diff
@@ -0,0 +1,57 @@
...
+        </span>
+        <span className="typography-body-small-14px-regular shrink-0 text-content-primary">
+          {formattedValue}
+        </span>
+      </div>
+      <ProgressBar
+        variant="sky"
+        size="medium"
+        value={value}
+        ariaLabel={typeof label === "string" ? label : undefined}
```
alfonsovo-fv, line 51: This silently loses the bar's name for anything that is not a string. `label` is typed `React.ReactNode`, so a translated element or a fragment makes this `undefined`, and `ProgressBar` then falls back to `aria-label={ariaLabel ?? "Progress"}` (line 162). On a stacked breakdown that is ten bars all announcing "Progress" with a number and nothing saying which country each one is.

The root of it is that the doc says "Pass a translated string" while the type allows any node, so the type permits the path that breaks. Two clean ways out:

Narrow the prop to `string`, which matches what the doc already asks for. Or point at the rendered label instead, which works for any node and avoids duplicating the text:

```tsx
const labelId = React.useId();
// on the label span: id={labelId}
<ProgressBar variant="sky" size="medium" value={value} aria-labelledby={labelId} />
```

The second needs `ProgressBar` to forward `aria-labelledby`, so the first is probably the cheaper fix here.

alfonsovo-fv (commented):
The `ProgressBar` additions are properly additive and the `muted-sky` token is correctly theme-aware, unlike the rest of the chart ramp, which is right for a background tint.

One real issue inline, plus the overlap with #641: this branch and `ds/country-flag` both carry `Demographics.tsx`, the `ProgressBar` changes and the `muted-sky` token, and both target `main`. #641's description says this one stacks on it, but the base here is `main`, so whichever merges first the other conflicts on the token files. Worth retargeting this onto `ds/country-flag`.

PR #643, opened 2026-07-31
fix(DropdownMenu): align row radius, radio padding and count size to V2

jacksonford-07:
Reopens #640, which GitHub auto-closed when its base branch `ds/dropdown-multiselect` was deleted on #639's merge. Same branch, rebased onto `main` with #639's commits dropped since they are squashed there now. **Already approved by @alfonsovo-fv on #640** — no changes since, so this needs no second review, only a merge.

Three pre-existing deviations from the design, none introduced by #639. Split out because each changes every dropdown in eden and fadmin, not just the rows #639 touched.

Every value is read from the Figma reference code, not inferred from a screenshot, and each is commented at the line with its node id.

| what | was | is | source |
| --- | --- | --- | --- |
| `DropdownMenuItem` row radius | `rounded-xs` (8px) | `rounded-sm` (12px) | `V2 Menu Item`, node `21542:8629` |
| `DropdownMenuRadioItem` padding | `px-4` (16px) | `px-3` (12px) | `V2 Menu Radio Item`, node `7393:62008` |
| count typography at size 40 | 16px/24px | 14px/18px | `V2 Menu Item`, node `21542:8629` |

**The two row radii genuinely differ, and that is not a mistake.** `V2 Menu Item` is `rounded-sm` (12px); `V2 Menu Radio Item` is `rounded-xs` (8px). So this changes the plain item and deliberately leaves the radio and checkbox items alone, with a test asserting the radio item *stays* at 8px so the next person to spot the inconsistency finds the reason instead of tidying it away.

**The count size compounded #639's alignment fix.** At 16px its line box is 24px against the two-line layout's 18px title leading, so it was both the wrong size and the wrong height. #639 centred it regardless, which had the effect of hiding the size error.

## Tests

Four, one per change plus the radio-radius guard. Verified against `main`: the three change assertions fail, the guard passes on both.

```
× gives a menu item the 12px row radius
✓ keeps the 8px radius on a radio item, which the design draws differently
× indents a radio item the same as any other row
× draws the count at 14px at both row heights
```

Verified in a local Storybook too, `FeatureItems` dark, every row's siblings sharing their title's centre:

```
row                     rowRadius  title        titleC  iconC  count        countC
Jane Doe (1 line)       12px       16px/24px    20.8    20.8   -            -
Alex Smith (2 line)     12px       14px/18px    17      -      14px/18px    17
Favourites (1 line)     12px       16px/24px    20      20     14px/18px    20
Edit profile (2 line)   12px       14px/18px    17      17     -            -
```

103 tests, `tsc --noEmit` clean.

## Not included, needs a decision

The header divider. Ours is full-bleed `-mx-2` with a comment saying the rule spans the panel deliberately; the search header in node `7393:62008` draws its `V2 Divider` with `pr-[4px]`. Left alone because the evidence is inconsistent across variants — the Default Desktop header (node `21549:6769`) has **no divider at all** and uses 12px padding where that instance uses 4px.

### DIFF
--- a/src/components/DropdownMenu/DropdownMenu.test.tsx
+++ b/src/components/DropdownMenu/DropdownMenu.test.tsx
@@ -83,6 +83,53 @@ describe("DropdownMenu", () => {
       expect(panel).not.toHaveClass("rounded-lg");
     });
 
+    // Each of these three drifted from the design unnoticed, so they are asserted
+    // rather than left to the eye. Node references are in the component.
+    it("gives a menu item the 12px row radius", () => {
+      renderMenu(<DropdownMenuItem>Item</DropdownMenuItem>);
+      const item = screen.getByRole("menuitem");
+      expect(item).toHaveClass("rounded-sm");
+      expect(item).not.toHaveClass("rounded-xs");
+    });
+
+    it("keeps the 8px radius on a radio item, which the design draws differently", () => {
+      renderMenu(
+        <DropdownMenuRadioGroup value="a">
+          <DropdownMenuRadioItem value="a">A</DropdownMenuRadioItem>
+        </DropdownMenuRadioGroup>,
+      );
+      const item = screen.getByRole("menuitemradio");
+      expect(item).toHaveClass("rounded-xs");
+      expect(item).not.toHaveClass("rounded-sm");
+    });
+
+    it("indents a radio item the same as any other row", () => {
+      renderMenu(
+        <DropdownMenuRadioGroup value="a">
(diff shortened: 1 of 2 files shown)


alfonsovo-fv (approved):
Approving. Same content as #640 rebased onto `main`, which I had already gone through, and the 88 DropdownMenu tests pass on the new base.

Re-stating the part worth keeping visible: the radio-radius guard is the valuable test here. An intentional inconsistency that nobody wrote down is exactly what the next person tidies away, and now they cannot without a test going red.

PR #644, opened 2026-08-01
feat(Tooltip, Select, ChartCard, Table): the DS changes the insights dashboard needs

jacksonford-07:
The insights dashboard was built against an unreleased `@fanvue/ui`, so Eden ended up carrying local overrides for things that belong in the library, and shipped without three props it was written to use. This closes that gap so the Eden side can drop its workarounds.

### The props

- **`ChartCard.skeleton`** — a placeholder shown in place of the children while `loading`, so a card keeps its shape instead of collapsing to just the header. Pass a `ChartSkeleton` of the matching variant. Omitting it keeps today's behaviour exactly.
- **`Table.size`** — for a table that sits in something other than a `TableCard` and so has no card to inherit density from. The insight cards put tables inside a `ChartCard`. Only provided to the cells when given, so omitting it still inherits.
- **`ChartTooltipContent.valueFormatter`** — changes a row's figure while keeping the row's layout, indicator and series name. A currency chart wants a price, not a thousands-separated number. `formatter` replaces the whole row and still takes precedence.

### Select

The Select popup and a `DropdownMenu` popup sit side by side in the insights toolbar and did not match — the Select panel was a flat background behind a plain shadow, the menu was the surface token behind the blur. Two `SelectItem` fixes ride along: the check indicator hung off the title's line on a two-line row, where the tick marks the whole row; and the row holding the current value painted identically to a merely hovered one, so it now takes the next step up the `neutral-alphas` ramp.

Both mirror what `DropdownMenuItem` already does. Its comment claimed `SelectItem` centred its tick too — that was aspirational, and is now true.

### Tooltip

Tooltips appeared and vanished instantly. They now fade, and take the v2 surface the design draws (tighter radius, visible border, lighter shadow).

Opacity only, deliberately: the content is portalled and positioned by Radix's popper, which owns its transform, and a transform here would also make it a stacking context. Declaring an exit animation is what keeps the content mounted long enough to play it.

### Blast radius

The tooltip change is the one to look at — **every tooltip in the product** changes shape and gains motion, and no automated check can catch that. Chromatic will show a wide diff. The Select surface affects every `Select` in Eden. The three props are additive and inert unless passed.

### Notes for review

- `valueFormatter` does **not** let Eden's `EarningsByTypeChart` drop its `formatter` override — that one needs the value-then-name row order, which is a separate change touching all 12 recharts cards. Left for its own PR.
- Rationale lives in JSDoc'd constants rather than inline comments, per `AGENTS.md` line 69.
- New stories cover the skeleton, the condensed table and the selected two-line row.

4666 unit tests pass, `tsc --noEmit` clean, no new lint warnings.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -69,6 +69,23 @@ export interface TooltipContentProps
   title?: React.ReactNode;
 }
 
+/**
+ * Fades the tooltip in and out, so it reads as arriving rather than blinking on.
+ *
+ * Opacity only, deliberately: the content is portalled and positioned by Radix's
+ * popper, which owns its transform — animating one here would fight it, and a
+ * transform on the content would also make it a stacking context. Declaring an
+ * exit animation is what keeps the content mounted long enough to play it, since
+ * Radix defers the unmount until `animationend`.
+ *
+ * Radix reports all three states through `data-state`, and the two open states
+ * are matched separately rather than as "not closed", so the exit rule can never
+ * be beaten by a broader selector. A consumer's own `data-[state=*]:` override
+ * still wins, since tailwind-merge sees the same variant and property.
+ */
+const ANIMATION_CLASSES =
+  "data-[state=delayed-open]:[animation:fv-tooltip-in_150ms_ease-out] data-[state=instant-open]:[animation:fv-tooltip-in_150ms_ease-out] data-[state=closed]:[animation:fv-tooltip-out_120ms_ease-in] motion-reduce:[animation:none]";
+
 export const TooltipContent = React.forwardRef<
   React.ComponentRef<typeof TooltipPrimitive.Content>,
   TooltipContentProps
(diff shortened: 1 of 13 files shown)


alfonsovo-fv (commented):
The three props being additive and inert unless passed makes this easy to take, and the Select fixes do mirror DropdownMenuItem. Two non-blocking notes plus one inline question:

- A few untouched story/test blocks got reflowed to multiline (ChartHelpers, Table.stories, Select.test, same thing in App.tsx on the scrim PR), looks like a narrower printWidth locally. Would be good to keep diffs to the real changes.
- Since every tooltip in the product changes shape, let's get design to eyeball the Chromatic build before we accept the 28 baselines.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -98,7 +115,8 @@ export const TooltipContent = React.forwardRef<
           collisionPadding={8}
           style={{ zIndex: "var(--fanvue-ui-portal-z-index, 50)", ...style }}
           className={cn(
-            "typography-description-12px-semibold max-w-[320px] rounded-sm bg-surface-primary-inverted px-4 py-2 text-content-primary-inverted shadow-[0px_1px_4px_0px_rgba(0,0,0,0.06),0px_1px_3px_0px_rgba(0,0,0,0.05)]",
+            "typography-description-12px-semibold max-w-[320px] rounded-xs border border-border-selected bg-surface-primary-inverted px-4 py-2 text-content-primary-inverted shadow-sm",
```
alfonsovo-fv, line 118: Is border-selected the token the design actually names here, or just the swatch that matched? A tooltip border riding the "selected" semantic means any future retheme of selection state restyles every tooltip with it. If Figma calls it something else it might be worth a dedicated alias even if it resolves to the same value today.

PR #645, opened 2026-08-03
feat(Logo): add the 3D brand icon via a `dimension` prop

jacksonford-07:
The v2 nav design uses a 3D treatment of the brand icon that `Logo` had no way to render. It's a genuinely different asset, not a recolour: depth comes from a radial lighten highlight over the green base, a gradient inner stroke, and a five-layer green shadow on the "F" which reads as an inner glow because it sits inside the clip path.

## Why a new prop rather than a new variant

`variant` is layout (full / icon / wordmark / portrait), `version` is sub-brand (default / agencies), and `color` is a scheme. The 3D art is none of those — it's a rendering of the same default mark — so it gets its own axis: `dimension="flat" | "3d"`, defaulting to `"flat"`. Nothing changes for existing callers.

Its colours are fixed, so it can't answer `color="decolour"` or the agencies sub-brand. Both take precedence and fall back to what they render today, mirroring how `useGlossyIcon` already handles decolour.

Happy to reshape the API if you'd rather fold this into `version` — flagging it as the one judgement call here.

## Implementation

Follows the `AgenciesIconSVG` precedent exactly: markup injection so the gradients, `mix-blend-mode` and filter survive intact, `isolate` so the blend doesn't reach the page behind it, and per-instance id namespacing so two logos on a page don't fight over the same gradient and filter defs.

## Verification

Rather than eyeball it, I diffed the committed asset against the Figma export programmatically — all 3 geometry paths byte-identical, all 6 gradient stops, the gradient coordinates, `gradientTransform`, `stroke-width` and all 10 filter colour matrices match.

Two notes for review:

- The export contains a duplicate of the "F" path at `fill-opacity=".01"`. It's invisible, and I kept it because `agenciesIcon.ts` keeps its equivalent — say the word and I'll strip both.
- The filter's declared bounds extend well past the 24x24 viewBox, so layers 3-5 are clipped away entirely. That's how the design renders it, so I've preserved it verbatim rather than pruning.

## Test plan

- [x] `pnpm exec vitest run src/components/Logo/` — 48 passing (24 unit + 24 storybook), including 4 new cases: 3D renders, decolour falls back to flat, agencies wins over 3D, and ids are namespaced across two instances
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm exec biome check src/` — clean (44 pre-existing warnings, none in Logo)
- [x] New `Icon3D` and `Full3D` stories for Chromatic to catch the render


alfonsovo-fv:
`dimension` as its own axis works for me, it's genuinely orthogonal to variant/version/color so I wouldn't fold it into version. And yes, say the word: I'd strip the invisible duplicate path here and in agenciesIcon in a follow-up, it's dead bytes in every consumer bundle.

PR #647, opened 2026-08-03
fix(CreatorTile): progressive-blur the bottom scrim (DES-815)

jacksonford-07:
Design QA on the in-chat creator callout asked for a progressive background blur over the tile's gradient, to add depth.

- Blur ramps 0 → 32px across the bottom 60%
- Gradient becomes `#151515` at 0% opacity from 40% of the height to 88% at the bottom

## Why five layers

`backdrop-filter` applies one uniform radius and `mask-image` then varies the alpha of the already-blurred result, so the single masked layer we had cross-faded a uniformly 12px-blurred copy over the sharp original — it faded contrast, not blur, and detail edges stayed visible to the bottom.

Sibling layers do ramp the radius: an element with `backdrop-filter` is a backdrop root for its *descendants*, so nesting samples nothing, while each later sibling samples everything already painted inside the root's `isolate`. Radii compound in quadrature, and the one-step overlap between adjacent trapezoid masks keeps the effective radius continuous rather than banded.

Verified in the browser that all of it compiles: 5 layers at `blur(2/4/8/16/32px)`, every mask present stepping 40% → 100%, gradient resolving to `#151515` at 88%, and `border-radius` inherited on each layer with the content row painting last and no ancestor filter.

## Notes for review

- Arbitrary `[mask-image:…]` rather than the `mask-b-from-*` utilities on purpose: those need Tailwind 4.1 and our peer range is `^4.0.0`, where they emit nothing and every layer would blur the whole tile at full radius.
- Layers are `inset-0` rather than `h-[60%]` so the mask percentages read directly against Figma's, and so each layer's edge-clamped `backdrop-filter` boundary sits where mask alpha is already 0.
- `rounded-[inherit]` not `rounded-b-[inherit]`, since the layers now have top corners on the tile boundary. Load-bearing on the blur layers: composited `backdrop-filter` descendants have a history of leaking square corners past an ancestor's `overflow-hidden` in WebKit.
- No `will-change` — 5 layers × 20 tiles in the Home carousel would trade a paint cost for a memory cliff.
- 88% is not on the `blackalpha` ramp (800/900/950 only) and no gradient token matches, so the primitive plus an alpha modifier carries it. A real token needs a Figma export; worth raising separately.
- Geometry improves rather than regresses: the scrim starting at 40% instead of 50% lowers the width at which the ~74px profile row rises above it for every preset (`banner` 370px → 308px). Documented in the `aspectRatio` JSDoc and shown by the new `NarrowBanner` story.
- **Heads-up for the two non-chat consumers:** `SuggestedCreatorsV2` and `SFWDiscoveryRecommendedCreators` both pass a `variant="primary"` Follow button, which is solid `#151515` — the same colour as the new scrim, at up to 88%. The pill keeps its white label but loses edge definition. The chat embed card is unaffected (`variant="white"`). Flagging for design rather than changing two surfaces that were not part of this QA pass.

Part of DES-815

### DIFF
--- a/src/components/CreatorTile/CreatorTile.tsx
+++ b/src/components/CreatorTile/CreatorTile.tsx
@@ -12,6 +12,38 @@ const ASPECT_RATIO_CLASSES: Record<CreatorTileAspectRatio, string> = {
   banner: "aspect-5/2",
 };
 
+/**
+ * Progressive blur over the scrim, ramping 0 to 32px across the bottom 60% of the tile.
+ *
+ * `backdrop-filter` applies one uniform radius and `mask-image` then varies the alpha of the
+ * already-blurred result, so a single masked layer cross-fades a uniformly blurred copy rather
+ * than ramping the radius. Sibling layers do ramp it: an element with `backdrop-filter` is a
+ * backdrop root for its descendants, so nesting samples nothing, while each later sibling samples
+ * everything painted before it inside the root's `isolate`. Radii therefore compound, in
+ * quadrature, and the one-step overlap between adjacent masks keeps the effective radius
+ * continuous instead of banding.
+ *
+ * Complete literal strings rather than interpolated radii, because Tailwind scans raw file text.
+ * Arbitrary `[mask-image:…]` rather than the `mask-b-from-*` utilities, which need Tailwind 4.1
+ * while the peer range is `^4.0.0` — on 4.0.x they emit nothing and every layer would blur the
+ * whole tile at full radius.
+ */
+const SCRIM_BLUR_LAYER_CLASSES = [
+  "pointer-events-none absolute inset-0 rounded-[inherit] backdrop-blur-[2px] [-webkit-mask-image:linear-gradient(to_bottom,transparent_40%,black_50%,black_60%,transparent_70%)] [mask-image:linear-gradient(to_bottom,transparent_40%,black_50%,black_60%,transparent_70%)]",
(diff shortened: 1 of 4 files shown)


alfonsovo-fv (commented):
Layering approach looks right to me, and the PR body answers most of what I'd have asked. One question inline about cost.

src/components/CreatorTile/CreatorTile.tsx
```diff
@@ -12,6 +12,38 @@ const ASPECT_RATIO_CLASSES: Record<CreatorTileAspectRatio, string> = {
...
+ * everything painted before it inside the root's `isolate`. Radii therefore compound, in
+ * quadrature, and the one-step overlap between adjacent masks keeps the effective radius
+ * continuous instead of banding.
+ *
+ * Complete literal strings rather than interpolated radii, because Tailwind scans raw file text.
+ * Arbitrary `[mask-image:…]` rather than the `mask-b-from-*` utilities, which need Tailwind 4.1
+ * while the peer range is `^4.0.0` — on 4.0.x they emit nothing and every layer would blur the
+ * whole tile at full radius.
+ */
+const SCRIM_BLUR_LAYER_CLASSES = [
```
alfonsovo-fv, line 31: Did we get a chance to scroll-test this on a real mid-range phone? We go from one half-height backdrop-filter per tile to five full-tile ones, so the Home carousel is compositing ~100 blur layers, and the masks don't reduce what each layer has to sample. If it chugs I think we could drop to 3 layers (4/11/32px still reads as a ramp) before touching the design.

PR #650, opened 2026-08-04
feat(AiDisclosureBadge): add EU AI disclosure badge

adampoulstonfv:
One of two from the EU AI Act disclosure set. Independent of #651, either order.

`AiDisclosureBadge` is Figma's `AI Disclosure Icon (EU)`: an "AI" wordmark on a filled disc, for overlaying AI-generated media. Eden carries its own copy of this today and it has already drifted, at 30% disc opacity where the spec says 50%.

**It is a component, not an icon.** The palette has to be fixed rather than theme-reactive, because the mark sits on arbitrary creator media and has to stay legible whatever is behind it. That rules out the `currentColor` contract the icon set is built on. Filed under `Icons/` it broke: `buttons-always-black-default` and `background-primary` are both `#151515` in dark theme, so the disc vanished into the page. Staying out of `Icons/` also keeps the diff off the generated test, story and barrel surface.

**`tone` is `dark | light | transparent`, not Figma's `Default | Negative | Transparent`.** `negative` already means something specific here — `Pill` and `Badge` both map it to `bg-buttons-secondary-negative-default`, backed by a real token family — so reusing it for the light treatment would collide. `dark` and `light` name the mark's own colour instead, which is also harder to misread. Figma should be renamed to match, or the next reader holds both vocabularies.

**`transparent` may not clear contrast, and that needs a design call.** It is the 50% black overlay token with white glyphs. Over near-white media it composites to about `#8a8a8a`, putting the wordmark at roughly 3.45:1 — fine against the 3:1 bar for a graphic, short of 4.5:1 if it counts as text. It holds above 4.5:1 while the media is darker than about `#d9d9d9`. The opacity is straight from the spec so I have not changed it, but bright media is the case this tone exists for.

Geometry is the Figma export run through the repo's own `scripts/svgo.config.mjs`, same as every icon.

The existing `AIDisclosureIcon` is a different mark (boxed glyph with sparkles, `currentColor`). The names are close enough that one of them may want renaming.

`Security Audit` fails on a pre-existing `fast-uri` advisory reached through devDependencies. No dependency changed here and it reproduces on `main`.

10 tests including axe, `tsc --noEmit` clean, biome clean.

### DIFF
--- a/src/components/AiDisclosureBadge/AiDisclosureBadge.test.tsx
+++ b/src/components/AiDisclosureBadge/AiDisclosureBadge.test.tsx
@@ -50,7 +50,7 @@ describe("AiDisclosureBadge", () => {
     it("keeps the palette fixed rather than following the inherited text colour", () => {
       const { container } = render(<AiDisclosureBadge />);
       for (const path of paths(container)) {
-        expect(path.getAttribute("class")).not.toContain("currentColor");
+        expect(path).not.toHaveClass("fill-current");
         expect(path).not.toHaveAttribute("fill");
       }
     });
--- a/src/components/AiDisclosureBadge/AiDisclosureBadge.test.tsx
+++ b/src/components/AiDisclosureBadge/AiDisclosureBadge.test.tsx
@@ -1,0 +1,77 @@
+import { render, screen } from "@testing-library/react";
+import { createRef } from "react";
+import { describe, expect, it } from "vitest";
+import { axe } from "vitest-axe";
+import { AiDisclosureBadge } from "./AiDisclosureBadge";
+
+describe("AiDisclosureBadge", () => {
+  const paths = (container: HTMLElement) => [...container.querySelectorAll("path")];
+
+  describe("API", () => {
+    it("applies custom className", () => {
+      const { container } = render(<AiDisclosureBadge className="custom-class" />);
+      expect(container.querySelector("svg")).toHaveClass("custom-class");
+    });
+
+    it("has a default size that className can override", () => {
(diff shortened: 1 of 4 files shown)


src/components/AiDisclosureBadge/AiDisclosureBadge.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    it("uses the half-opacity overlay fill for the transparent tone", () => {
+      const [disc, glyph] = paths(render(<AiDisclosureBadge tone="transparent" />).container);
+      expect(disc).toHaveClass("fill-background-overlay-default");
+      expect(glyph).toHaveClass("fill-content-always-white");
+    });
+
+    it("keeps the palette fixed rather than following the inherited text colour", () => {
+      const { container } = render(<AiDisclosureBadge />);
+      for (const path of paths(container)) {
+        expect(path.getAttribute("class")).not.toContain("currentColor");
```
alfonsovo-fv: I think this assertion can't fail: a Tailwind class for inherited colour would be `fill-current`, not the literal string "currentColor", so this passes no matter what. The `not.toHaveAttribute("fill")` line below is the one doing the work. Maybe check for `fill-current` instead, or drop this line?

alfonsovo-fv (approved):
One test nit inline, otherwise this is great. The component-not-icon call and staying off `negative` both make sense to me, and I agree the transparent contrast question belongs with design rather than being quietly changed here.

alfonsovo-fv (approved):

PR #651, opened 2026-08-04
feat(AiDisclosureLabel): add EU AI generated/modified disclosure label

adampoulstonfv:
The other half of the EU AI Act disclosure set. Independent of #650, either order.

`AiDisclosureLabel` is Figma's `AI Modified Label (EU)`: an "AI" mark followed by wording, on a filled pill.

**The wording ships as live text, not Figma's outlines.** Figma draws it as vector paths and enumerates the two English wordings as variants. Outlines would be pixel-exact and permanently English: no translation, no scaling with the reader's font settings, nothing for a screen reader. So `label` is a required string the caller passes already translated, uppercased in CSS so callers pass sentence case rather than shouting. The cost is that it is not pixel-identical to the frame. There is deliberately no `kind` prop supplying English defaults, since a design system shipping those is how untranslated strings reach production.

**`tone` is `dark | light | transparent`, not Figma's `Default | Negative | Transparent`** — same reason as #650: `negative` already maps to `bg-buttons-secondary-negative-default` in `Pill` and `Badge`. Figma should be renamed to match.

**Two typography details are load-bearing.** The pill layers `typography-description-12px-semibold` under per-run `text-[Npx]` sizes, because those utilities are the only place this library sets `font-family: Inter` and without it the mark fell back to a system font. `Count` layers the same way for its sub-12px steps. Letter-spacing sits on each text run as an `em` rather than on the pill, so it scales with `size`; on the pill it would resolve once against the inherited font size and freeze at one absolute value for all three.

**`role="img"` with a composed accessible name.** "AI" and the wording are separate elements with only a flex gap between them, so without a name of its own the mark announces as the single token "AIGenerated".

Three for design, all implemented as drawn rather than quietly diverged: `transparent` puts white text at about 3.45:1 over near-white media, under the 4.5:1 text bar (it clears it while media is darker than roughly `#d9d9d9`, and the 50% opacity is from the spec); at `size="12"` the wording renders at 6px, which does not clip but is a legibility question; and both runs are weight 600, though the Figma render reads as though the "AI" is heavier — not recoverable from outlined vectors, so say the word and I will split it.

`Badge` already has an `aiGenerated` variant whose classes are exactly this component's `light` tone. Worth deciding whether a spec-driven disclosure mark and general AI tagging converge or stay separate.

`Security Audit` fails on a pre-existing `fast-uri` advisory reached through devDependencies. No dependency changed here and it reproduces on `main`.

22 tests including axe, `tsc --noEmit` clean, biome clean.

### DIFF
--- a/src/components/AiDisclosureLabel/AiDisclosureLabel.stories.tsx
+++ b/src/components/AiDisclosureLabel/AiDisclosureLabel.stories.tsx
@@ -1,0 +1,59 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import { AiDisclosureLabel } from "./AiDisclosureLabel";
+
+const meta = {
+  title: "Components/AiDisclosureLabel",
+  component: AiDisclosureLabel,
+  parameters: {
+    layout: "centered",
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/S8zFdcOjt4qN4PrwntuCdt/Fanvue-Library?node-id=20586-1134&m=dev",
+    },
+  },
+  tags: ["autodocs"],
+  args: { label: "Generated" },
+  argTypes: {
+    size: { control: "inline-radio", options: ["12", "16", "20"] },
+    tone: {
+      control: "inline-radio",
+      options: ["dark", "light", "transparent"],
+    },
+  },
+} satisfies Meta<typeof AiDisclosureLabel>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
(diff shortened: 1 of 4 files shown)


alfonsovo-fv (approved):
Looks good. Live text over outlines is the right call, and the typography layering works by inheritance so there is no tailwind-merge fragility. The three design questions in the description read as design decisions, not blockers. Would be good to get an answer on the Badge aiGenerated convergence at some point, but that should not hold this up.

PR #652, opened 2026-08-04
fix(AnimatedNumber): key roll columns by place value, box the separators

jacksonford-07:
Two bugs in `AnimatedNumber`'s `roll` variant, both plain to see on a headline money figure — a 40px `Header/Heading lg` total that re-fetches when a date range or a Net/Gross filter changes. Same component, same render, so they come together.

## The number snapped instead of rolling

A column only animates if its DOM node survives the re-render and its inline `translateY` changes; there is no from-value, so a freshly mounted column paints at its final offset with no transition.

Columns were keyed positionally from the **left** of the formatted string, with separate prefixes for the two kinds (`digit-${index}` / `separator-${index}`). Any change in the number of characters therefore renumbered every position. `$12,345.67` → `$1,234.56` is length 10 → 9, so `digit-2` becomes `separator-2`, `separator-3` becomes `digit-3`, and most columns are destroyed and recreated. Only `boxWidth` survives, because it is component state — which is exactly the "box resizes smoothly but the digits jump" look.

Keyed by distance from the **end**, a column is a place value. Counting from the least-significant end is how an odometer behaves and what keeps money aligned: the decimal separator, the decimal places and the thousands separators all sit at stable offsets from the right, so every shared position keeps its identity and rolls, and only new leading digits mount.

Worth noting this was never only about changing magnitude by an order — any value change that crosses a digit-count boundary snapped, including a Net/Gross toggle from `$1,024.37` to `$998.12`. General correctness, not one filter's edge case.

## Every `$`, `,` and `.` sat above the digits

The digit columns set `height: 1em` explicitly; the separator spans set nothing. So on a flex line whose only stretchable child is a separator, the separator took the full inherited line-height — 44px against the intended 40px at `Header/Heading lg` — and put its glyph on a different baseline from the figure.

Two consequences, one visible immediately and one further out:

- the currency symbol and both separators rode high inside the number
- the component's own box grew 4px taller than its type. `RollingNumber`'s outer span is `inline-flex overflow-hidden`, and an overflow-hidden inline box reports its **bottom margin edge** as its baseline — so anything a consumer baseline-aligns beside the number was measured against an inflated box and landed well below the digits. Negligible at `ChartCard`'s 24px, obvious at 40px.

Giving the separator the same `1em` box fixes both.

## Test

`keeps a place value's column across a change of magnitude, so it rolls` — renders `999`, grabs the units column, re-renders at `1000`, and asserts the **same DOM node** is still there at the new index with `translateY(-0%)`. Node identity is the thing the transition needs, and the existing suite only covered first paint, so nothing caught this.

Verified it discriminates: with the old left-anchored keys that test fails (`after[3]` is a fresh node) and the other 11 still pass.

## Notes for review

- **No API change**, no new props, nothing to adopt. Consumers on `variant="roll"` get a roll that works; `variant="count"` is untouched.
- Both `biome-ignore lint/suspicious/noArrayIndexKey` suppressions are gone. The keys now derive from a named offset rather than the map's `index` parameter, so the rule no longer fires and biome reported both as unused. The reasoning they carried — a column is a place value, so it must keep its identity across a digit change — is in the comment above the map, where it now also has to explain the direction.
- First paint is byte-identical; only the keys and the separator's `style` differ, so there is no snapshot or visual-regression fallout to expect.
- The separator fix is the reason I am **not** sending a follow-up to give the roll variant a synthesized text baseline. That would still be a real improvement (the box's bottom edge is a descender below the glyphs even at the correct height), but it needs the digit strip to stop relying on each digit being exactly 1/10 of a 1em window, and it is no longer blocking anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
       "lodash-es": ">=4.18.0",
       "minimatch": ">=10.2.3",
       "ajv": ">=8.18.0",
-      "fast-uri": ">=3.1.4",
+      "fast-uri": ">=4.1.2",
       "brace-expansion": ">=5.0.9",
       "yaml": ">=2.8.3",
       "postcss": ">=8.5.18",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,7 +9,7 @@ overrides:
   lodash-es: '>=4.18.0'
   minimatch: '>=10.2.3'
   ajv: '>=8.18.0'
-  fast-uri: '>=3.1.4'
+  fast-uri: '>=4.1.2'
   brace-expansion: '>=5.0.9'
   yaml: '>=2.8.3'
   postcss: '>=8.5.18'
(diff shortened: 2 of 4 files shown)


jacksonford-07:
The second commit here bumps the `fast-uri` override and is unrelated to this change — review this PR on its first commit.

`pnpm audit --audit-level=high` reads the live advisory database, so [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) being published yesterday turned `Security Audit` red on every fresh CI run in the repo, #651 included. Our override floor of `>=3.1.4` was satisfied by the vulnerable `4.1.1`, so it now reads `>=4.1.2`. The audit gate itself is untouched: no ignore list, no lowered level.

The same commit is on #653 (and #652). Identical content, so whichever merges first makes the other a no-op with no conflict — and until one of them lands, other people's PRs keep failing that check through no fault of their own.

alfonsovo-fv (approved):
Diagnosis and fix both look right, and keying from the least significant end holds up for locales that trail the currency symbol too. Noted your comment on the fast-uri commit.

PR #653, opened 2026-08-04
feat(Chart): add ChartAreaGradientDefs for area fills

jacksonford-07:
One vertical gradient per series for the area fills on a line or area chart: the series colour where it meets its line, fading to nothing at the baseline. `ChartAreaGradientDefs` emits the `<defs>`, `chartAreaGradientFill(idPrefix, key)` addresses one.

## Why not a flat fill

A flat translucent fill is what this replaces. Two of them overlapping read as a third colour and the eye cannot tell which series is in front — a fade keeps each series legible where they cross and puts the visual weight on the line, which is where the value is.

The `AreaGradients` story renders two series that cross three times, which is the case that motivated it.

## Why it belongs in the DS

`AreaChart.stories.tsx` already hand-rolls exactly this, through a story-only `AreaGradient` fixture in `chartStoryFixtures.tsx`, at seven call sites with hardcoded ids (`fillEarnings`, `fillDesktop`, `fillMobile`, …). Eden carries a second copy for the insights earnings chart. Same three-line `<linearGradient>` in three places, none of them shared.

## Design notes

- **Colours resolve through `var(--color-${key})`**, the custom property `ChartStyle` already publishes from the chart config, rather than taking a colour prop. A gradient therefore tracks whatever colour that series is given, including its dark-theme override, with no second place to keep in sync.
- **`idPrefix` is required, not generated.** SVG ids are document-global and several of these charts appear on one page, so the prefix has to come from the caller's `useId()`. Generating one inside would either need the component to own the fill lookup too or hand back an id the caller cannot reconstruct.
- **The defs render inside the chart**, not beside it. `url(#…)` resolves against the document, but recharts will unmount anything outside its own SVG subtree.
- **`topOpacity` defaults to 0.3**, taken from the V2 insights frames. The story fixture's 0.8 → 0.1 is a recharts-example value, not a design one; the prop is there for anyone who genuinely needs a heavier fill (`AreaGradientsStrongerFill` shows 0.6).
- **Do not also set `fillOpacity`** on the `Area` — it multiplies the gradient's own alpha and flattens the fade. Documented in the JSDoc.

## Notes for review

- **No `ref`, no `className`.** It emits `<defs>`; there is no box to style and no host element a consumer would reach for. Same shape as `ChartStyle`, which is also a plain function component with neither.
- Exported from `charts.ts` only, alongside the other chart helpers — not from `src/index.ts`, matching how the chart entrypoint is split today.
- One test carries a comment worth knowing about: jsdom lowercases a camelCase type selector when it appears in a combinator, so `querySelectorAll("linearGradient stop")` matches **nothing** while `querySelectorAll("linearGradient")` matches fine. Stops are read off the gradient node instead. I lost a few minutes to that; the next person doesn't have to.
- **Deliberately not converting `AreaChart.stories.tsx`** to use this. Those seven call sites would change appearance (0.8 → 0.1 becomes 0.3 → 0), so it is a visual review of six stories rather than a mechanical swap, and the story fixture can go in the same follow-up.
- 129 chart tests pass, `tsc` clean, and `biome check .` reports the same 42 pre-existing warnings as `main` — none added.

### DIFF
--- a/src/components/Chart/ChartAreaGradient.tsx
+++ b/src/components/Chart/ChartAreaGradient.tsx
@@ -1,0 +1,73 @@
+/** Props for {@link ChartAreaGradientDefs}. */
+export interface ChartAreaGradientDefsProps {
+  /**
+   * Namespace for the generated gradient ids. SVG ids are document-global, so this
+   * has to be unique per chart on the page — pass a `React.useId()`.
+   */
+  idPrefix: string;
+  /**
+   * Series to emit a gradient for. Each resolves `var(--color-<key>)`, the custom
+   * property {@link ChartStyle} already publishes from the chart config, so the
+   * gradient tracks whatever colour that series is given.
+   */
+  seriesKeys: readonly string[];
+  /**
+   * Opacity where the fill meets the line, fading to nothing at the baseline.
+   * @default 0.3
+   */
+  topOpacity?: number;
+}
+
+/** The `fill` value for a series' gradient, as emitted by {@link ChartAreaGradientDefs}. */
+export const chartAreaGradientFill = (idPrefix: string, seriesKey: string) =>
+  `url(#${idPrefix}-${seriesKey})`;
+
+/**
+ * One vertical gradient per series, for the area fills on a line or area chart:
+ * the series colour where it meets its line, fading out towards the baseline.
+ *
+ * Render it inside the chart — SVG resolves `url(#…)` against the document, but the
+ * `defs` must be part of the chart's own SVG for recharts to keep it mounted — and
+ * fill each `Area` with {@link chartAreaGradientFill}. Do not also set
+ * `fillOpacity`: it multiplies the gradient's own alpha and flattens it.
+ *
+ * A flat translucent fill is what this replaces. Two of them overlapping read as a
+ * third colour, and the eye cannot tell which series is in front; a fade keeps each
+ * series legible where they cross and puts the weight on the line.
+ *
+ * @example
+ * ```tsx
+ * const gradientId = React.useId();
+ *
+ * <AreaChart data={rows}>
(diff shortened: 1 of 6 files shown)


jacksonford-07:
The second commit here bumps the `fast-uri` override and is unrelated to this change — review this PR on its first commit.

`pnpm audit --audit-level=high` reads the live advisory database, so [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) being published yesterday turned `Security Audit` red on every fresh CI run in the repo, #651 included. Our override floor of `>=3.1.4` was satisfied by the vulnerable `4.1.1`, so it now reads `>=4.1.2`. The audit gate itself is untouched: no ignore list, no lowered level.

The same commit is on #653 (and #652). Identical content, so whichever merges first makes the other a no-op with no conflict — and until one of them lands, other people's PRs keep failing that check through no fault of their own.

src/components/Chart/ChartAreaGradient.tsx
```diff
@@ -0,0 +1,73 @@
+/** Props for {@link ChartAreaGradientDefs}. */
+export interface ChartAreaGradientDefsProps {
+  /**
+   * Namespace for the generated gradient ids. SVG ids are document-global, so this
+   * has to be unique per chart on the page — pass a `React.useId()`.
```
alfonsovo-fv, line 5: The JSDoc says to pass a React.useId(), but that comes back with colons and both ChartMetricTrend and the Eden copy of this strip them before use. Could the component strip them itself? Otherwise every caller has to remember, and forgetting only bites in a selector rather than in url(#...).

alfonsovo-fv (commented):
Gradient approach makes sense and the reasoning for it over a flat fill is convincing. One question on the id handling.

PR #654, opened 2026-08-04
fix(deps): raise the fast-uri override past GHSA-7p8r-x3mc-p8w7

jacksonford-07:
`Security Audit` is failing on every PR opened since yesterday, including #651, #652 and #653. It is not any of those changes — `pnpm audit --audit-level=high` queries the live advisory database, so a newly published advisory turns previously-green CI red with no code change. Main's last CI run was the 3.25.0 release on Aug 3, before it landed.

## The advisory

[GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) — `fast-uri` host confusion via a backslash authority introducer. **High.** Vulnerable `>=4.0.0 <4.1.2`, patched in `4.1.2`.

We were on `4.1.1`, and the reason is the override itself:

```
"fast-uri": ">=3.1.4"
```

That floor is satisfied by `4.1.1`, so the override that exists to keep this package patched was pinning us **inside** the vulnerable range. Raised to `>=4.1.2`, which is the same remedy the rest of that list already applies (`minimatch >=10.2.3`, `brace-expansion >=5.0.9`, and so on).

`pnpm audit --audit-level=high` now exits 0.

## Notes for review

- **Two files, six lines.** `fast-uri` reaches us only through `ajv`, under `@commitlint/cli` and `vite-plugin-dts > @microsoft/api-extractor`, so nothing in the published bundle changes — 16 paths, all dev tooling.
- Verified locally after a real install: `pnpm build` succeeds, `tsc --noEmit` clean, `biome check .` reports the same 42 pre-existing warnings as main, and all 3,657 unit tests pass.
- **The audit gate itself is untouched.** No ignore list, no lowered `--audit-level` — the vulnerable version is gone instead.
- One **moderate** finding remains and is worth a follow-up on the same pattern: `postcss` is overridden at `>=8.5.18` while GHSA-6g55-p6wh-862q's incomplete-fix advisory wants `>=8.5.23`. It does not fail the `high` gate, and postcss sits in the Tailwind build path, so it deserves its own PR and its own build check rather than riding along with an unblocking fix.
- Once this merges, #652 and #653 need their branches updated to pick it up before they can go green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
       "lodash-es": ">=4.18.0",
       "minimatch": ">=10.2.3",
       "ajv": ">=8.18.0",
-      "fast-uri": ">=3.1.4",
+      "fast-uri": ">=4.1.2",
       "brace-expansion": ">=5.0.9",
       "yaml": ">=2.8.3",
       "postcss": ">=8.5.18",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,7 +9,7 @@ overrides:
   lodash-es: '>=4.18.0'
   minimatch: '>=10.2.3'
   ajv: '>=8.18.0'
-  fast-uri: '>=3.1.4'
+  fast-uri: '>=4.1.2'
   brace-expansion: '>=5.0.9'
   yaml: '>=2.8.3'
   postcss: '>=8.5.18'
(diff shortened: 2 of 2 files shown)


jacksonford-07:
Closing — the same two-file change now rides in #652 and #653, so it reaches main with whichever of those merges first and does not need a PR of its own.

Heads-up for anyone who lands here from a red `Security Audit`: the fix is bumping the `fast-uri` override from `>=3.1.4` (satisfied by the vulnerable 4.1.1) to `>=4.1.2`, for [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7). Until #652 or #653 merges, other PRs will keep failing that check through no fault of their own.

